PR #41, opened 2015-01-11
add test for description after .execute

jettify:
I've added test for ``cursor.description`` after ``cursor.execute()``
Please review.


asvetlov:
Cool. Thanks

PR #42, opened 2015-01-16
issue with cursor __iter__

jettify:
I've modified little bit ``test_iter(self)`` test:

```python
    def test_iter(self):
        @asyncio.coroutine
        def go():
            conn = yield from self.connect()
            cur = yield from conn.cursor()
            yield from cur.execute("SELECT * FROM tbl")
            data = [(1, 'a'), (2, 'b'), (3, 'c')]
            fetched_data = []
            for item in cur:
                fetched_data.append(item)
            self.assertEquals(fetched_data, data)
```
And following issue poped up.

```python

======================================================================
FAIL: test_iter (tests.test_cursor.TestCursor)
----------------------------------------------------------------------
  File "tests/test_cursor.py", line 449, in test_iter
  File "/usr/lib/python3.4/asyncio/base_events.py", line 276, in run_until_complete
    return future.result()
  File "/usr/lib/python3.4/asyncio/futures.py", line 277, in result
    raise self._exception
  File "/usr/lib/python3.4/asyncio/tasks.py", line 237, in _step
    result = next(coro)
  File "tests/test_cursor.py", line 448, in go
    self.assertEquals(fetched_data, data)
AssertionError: Lists differ: [] != [(1, 'a'), (2, 'b'), (3, 'c')]

Second list contains 3 additional elements.
First extra element 0:
(1, 'a')

- []
+ [(1, 'a'), (2, 'b'), (3, 'c')]
```

looks like ``yield from``  or ``raise StopIteration``in  ``Cursor.__iter__`` method kills iteration.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,9 @@
 CHANGES
 -------
+0.6.1
+^^^^^
+
+* Drop iteration protocol support in cursor.
 
 0.6.0 (2015-02-03)
 ^^^^^^^^^^^^^^^^^^
--- a/aiopg/cursor.py
+++ b/aiopg/cursor.py
@@ -360,10 +360,3 @@ class Cursor:
     def timeout(self):
         """Return default timeout for cursor operations."""
         return self._timeout
-
-    def __iter__(self):
-        item = yield from self.fetchone()
-        if item is None:
-            raise StopIteration
-        else:
-            return item
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -26,7 +26,7 @@ def get_release():
             if match is not None:
                 return match.group(1)
         else:
-            raise RuntimeError('Cannot find version in aiozmq/__init__.py')
+            raise RuntimeError('Cannot find version in aiopg/__init__.py')
 
 
 def get_version(release):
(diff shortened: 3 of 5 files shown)


asvetlov:
Ooh. Perhaps adding `__iter__` method to `cursor` was a mistake.
Taras Voinarovsky pointed me on the problem a months ago.
Looks like I just have to drop `__iter__` method but not ready to do it just now -- please wait for next week.

I need to think and to talk with all you both.

jettify:
I've removed ``__iter__`` method completely just to see any dependency to it. Looks like we can delete it painlessly.

asvetlov:
Would you fix the docs also?

I see `_cursor-iterable` section at least in the `./docs/core.rst`.

We should replace it with `.. warning::`  about `__iter__` removing with strong explanation why we have to do the change.

Also CHANGES.txt should be updated by adding next version 0.6.0 with the same note.

I recall the request to add iteration over cursor, our users should be notified about dropping the way.

jettify:
sure, but bit later

jettify:
I've update docs, please review.

asvetlov:
Thanks

asvetlov:
Hmm. Should https://github.com/aio-libs/aiopg/blob/master/aiopg/sa/result.py#L321 be removed also?

PR #47, opened 2015-02-06
add coveralls and badges

jettify:
I've added coveralls to travis config, you just need to enable this repository in coveralls panel.


asvetlov:
Thanks!

PR #52, opened 2015-04-03
Add psycopg2cffi support

GMLudo:
Hi,

This pull request adds `psycopg2cffi` support for `aiopg`.
Strategy used: if `psycopg2cffi` is present, use that instead of `psycopg2`.
If you have a better strategy to handle this use case without to be overcomplicated, feel free to propose something else.

Almost suite tests works, but you have small differences between both, like for dsn representation:

```
FAIL: test_dsn (tests.test_connection.TestConnection)
----------------------------------------------------------------------
File "tests/test_connection.py", line 197, in test_dsn
File "/home/lg/.pythonz/pythons/CPython-3.4.3/lib/python3.4/asyncio/base_events.py", line 316, in run_until_complete
    return future.result()
File "/home/lg/.pythonz/pythons/CPython-3.4.3/lib/python3.4/asyncio/futures.py", line 275, in result
    raise self._exception
File "/home/lg/.pythonz/pythons/CPython-3.4.3/lib/python3.4/asyncio/tasks.py", line 238, in _step
    result = next(coro)
File "tests/test_connection.py", line 195, in go
    conn.dsn)
AssertionError: 'dbname=aiopg user=aiopg password=xxxxxx host=127.0.0.1' != 'dbname=aiopg user=aiopg password=passwd host=127.0.0.1'
- dbname=aiopg user=aiopg password=xxxxxx host=127.0.0.1
?                                  ^^^^^^
+ dbname=aiopg user=aiopg password=passwd host=127.0.0.1
?                                  ^^^^^^
```

I imagine it isn't intentional, the goal of `psycopg2cffi` is to replace `psycopg2`.

From performance perspective, I've launched quickly some benchmarks with CPython 3.4.3: At least with the tests I've used from FrameworkBenchmarks, with `psycopg2`, it's always `psycopg2` that wins (+/- 30 %).

Nevertheless, integrate `psycopg2cffi` as alternative has two benefits:

With `psycopg2`, when one of my daemon is under high load, I've this error several times in my syslog:

```
asyncio:ERROR 2015-04-03 22:37:31,189 base_events.py:912 => Fatal error on aiopg connection: bad state in _ready callback connection: <aiopg.connection.Connection object at 0x7f0b6e938dd8>
```

With `psycopg2cffi`, no errors.

Moreover, with `psycopg2cffi` support, it's possible to use PyPy3 for `aiopg`. It isn't really functional for now, but I don't want to maintain an `aiopg` fork only for `psycopg2cffi` support during to try improve PyPy3 support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ build
 .coverage
 htmlcov
 docs/_build
+pyvenv*/
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .idea
+*.iml
 *.pyc
 __pycache__
 dist
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -1,9 +1,15 @@
 import asyncio
 
-import psycopg2
-from psycopg2.extensions import (
-    POLL_OK, POLL_READ, POLL_WRITE, POLL_ERROR)
-from psycopg2 import extras
+try:
+    import psycopg2cffi as psycopg2
+    from psycopg2cffi.extensions import (
+        POLL_OK, POLL_READ, POLL_WRITE, POLL_ERROR)
+    from psycopg2cffi import extras
+except ImportError:
+    import psycopg2
+    from psycopg2.extensions import (
+        POLL_OK, POLL_READ, POLL_WRITE, POLL_ERROR)
+    from psycopg2 import extras
 
 from .cursor import Cursor
 
--- a/aiopg/cursor.py
+++ b/aiopg/cursor.py
@@ -1,6 +1,9 @@
 import asyncio
 
-import psycopg2
+try:
+    import psycopg2cffi as psycopg2
+except ImportError:
+    import psycopg2
 
 from .log import logger
 
--- a/aiopg/pool.py
+++ b/aiopg/pool.py
@@ -1,8 +1,10 @@
 import asyncio
 import collections
 
-
-from psycopg2.extensions import TRANSACTION_STATUS_IDLE
+try:
+    from psycopg2cffi.extensions import TRANSACTION_STATUS_IDLE
+except ImportError:
+    from psycopg2.extensions import TRANSACTION_STATUS_IDLE
 
 from .connection import connect, TIMEOUT
 from .log import logger
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -4,10 +4,11 @@
    contain the root `toctree` directive.
 
 aiopg
-=================================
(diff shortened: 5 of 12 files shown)


GMLudo:
I've forgotten to say that this `aiopg` version runs only with `psycopg2cffi` from Git master branch, not with the latest PyPI release.
I've opened an issue about that on the `psycopg2cffi` bug tracker: https://github.com/chtd/psycopg2cffi/issues/35

asvetlov:
I like the idea in general but tests should cover both underlying libraries.

asvetlov:
Another question is: what version should be default?
In your path ``psycopg2cffi`` is preferred, don't sure this is good.
Anyway fallback should be forced by environment variable.

GMLudo:
OK for the tests, but `psycopg2cffi` differences with `psycopg2` should be fixed before to enable tests in `aiopg`. I'll see that with `psycopg2cffi` developers first.

For `psycopg2cffi` or `psycopg2` selection, I've taken the simplest solution.
From my point of view, this choice should be given explicitly when you instantiate an `aiopg` Pool or an `aiopg` Connection, like you did in `aiohttp`.
But for `aiohttp` it was more simple because you pass class definitions as `__init__()` parameter.
Here, it's more complicated because it's a Python package. Nevertheless, it should be possible to use `importlib` to do that.
Environment variable is also a solution, less visible that an explicit parameter, but cleaner that to need to install or uninstall `psycopg2cffi` to choose.

`psycopg2cffi` has also a magical function to switch a project from `psycopg2` to `psycopg2cffi` that you need to launch before to import your application:

    from psycopg2cffi import compat
    compat.register()

I understand why they added that: to facilitate migrations.
But, at least for me, it's like implicit async behavior you have with gevent: If you don't spot that in fact you're in a gevent code instead of standard python code, you can have behavior differences.

Maybe I'll ask on Python ML if somebody else has also implemented a mechanism to choose a Python package use in a library. Maybe another solution should exist.

kxepal:
> Maybe I'll ask on Python ML if somebody else has also implemented a mechanism to choose a Python package use in a library. Maybe another solution should exist.

I know one of the such: https://github.com/djc/couchdb-python/blob/master/couchdb/json.py
Not ideal, but does well what it should.

jettify:
Branch is very outdated, closing for now, feel free to open new updated PR

GMLudo:
I've an updated version with the latest PyPy3, not yet released.

I'll open again if I've something to show.

PR #54, opened 2015-04-17
support for sqlalchemy 1.0.0

jettify:
SQLAchemy's 1.0.0 release is not working with current aiopg due to change of internal api. I do not know proper way to fix but changing `result_map` to `_create_result_map()` works fine.


jettify:
better fix in #55

PR #56, opened 2015-04-23
Added test for default values on insert (Test first)

eirnym:
Defaults are not supported for now

### DIFF
--- a/aiopg/sa/utils.py
+++ b/aiopg/sa/utils.py
@@ -1,0 +1,117 @@
+import asyncio
+from functools import wraps
+from io import StringIO
+
+import sqlalchemy
+
+
+def dump_sql(func, bind=False, dialect=None):
+    @wraps(func)
+    def func_wrapper(*args, **kwargs):
+        out = StringIO()
+
+        def dump(sql, *multiparams, **params):
+            out.write('{};\n'.format(str(
+                sql.compile(dialect=dump.dialect)).strip()))
+
+        # There's no way to use given dialect here
+        engine = sqlalchemy.create_engine('postgres://',
+                                          strategy='mock',
+                                          executor=dump)
+        if dialect is None:
+            dump.dialect = engine.dialect
+        else:
+            dump.dialect = dialect
+
+        if bind:
+            func(*args, bind=engine, **kwargs)
+        else:
+            func(engine, *args, **kwargs)
+
+        return out.getvalue()
+    return func_wrapper
+
+
+def create_all_sql(metadata, dialect=None):
+    return dump_sql(metadata.create_all, bind=True, dialect=dialect)()
+
+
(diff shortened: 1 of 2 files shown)


eirnym:
Sample working code to help to fix this test:

    @asyncio.coroutine
    def get_column_default(engine, column):
        default = column.default
        dialect = engine.dialect
        if default.is_sequence:
            sql = "select nextval('%s')" % \
                dialect.identifier_preparer.format_sequence(default)
            print(sql)
            with (yield from engine) as conn:
               return (yield from conn.scalar(sql))
        elif default.is_clause_element:
            with (yield from engine) as conn:
               return (yield from conn.scalar(default.arg))
        elif default.is_callable:
            return default.arg(dialect)
        else:
            return default.arg

    @asyncio.coroutine
    def get_table_defaults(engine, table):
        result = {}
        for column in table.c:
            if column.default is not None:
                result[column.name] = (yield from get_column_default(engine, column))
        return result

    @asyncio.coroutine
    def insert(engine, table, returning_id=False, **kwargs):
        values = yield from get_table_defaults(engine, table)
        values.update(kwargs)
        print(values)
        if returning_id:
            insertion = table.insert(returning=[table.id])
        else:
            insertion = table.insert()
        return insertion.values(**values)

eirnym:
But better way is to implement somehow ExecutionContext if it possible with mixing async and not async code.

eirnym:
Added test for this override

asvetlov:
@eirnym I don't know how to hack ExecutionContext but maybe calculating defaults in `SAConnection.execute()` may help?
See https://github.com/aio-libs/aiopg/blob/master/aiopg/sa/connection.py#L71 for the modification point  what I guess.

eirnym:
Test `test_defaults_override `, uses function which extracts defaults from table and uses almost same code as in SQLAlchemy . I'll try do same with onupdate later.

eirnym:
@asvetlov It's not too easy to put this into connection. `Connection` object executes `dialect` first, then dialect executes `connection._cursor_execute`, which executes `dialect.dispatch` methods for events, defaults and so on.

So version with fabric is more stable and understandable for others than that *hack* which is used in `SQLAlchemy`.

eirnym:
In other words, it's easer to rewrite entire `SQLAlchemy` to use `asyncio` than inject defaults into `Connection` object.

asvetlov:
If there is impossible to use native SQLAlchemy defaults processing in our code we should document it explicitly to encourage people using our versions of `insert` and `update` functions.

Maybe we still can support default like constant and callable for native queries and force user to use our functions for sequences and clauses?

vir-mir:
I close, problem with default value for column resolved here -> (https://github.com/aio-libs/aiopg/pull/206)

PR #59, opened 2015-05-19
Support pools with unlimited size

aplavin:
According to documentation, maxsize = 0 means unlimited.


asvetlov:
Please add at least single test.

Issue #68, opened 2015-08-18
SQLAlchemy example doesn't work to me


hyzhak:
This one example https://github.com/aio-libs/aiopg#example-of-sqlalchemy-optional-integration doesn't work. And gives error: 

```
psycopg2.ProgrammingError: relation "tbl" does not exist
LINE 1: INSERT INTO tbl (val) VALUES ('abc') RETURNING tbl.id
```

Because table `tbl` isn't created by default. And as I understand we should add something like: `yield from conn.execute(CreateTable(tbl))` before `yield from conn.execute(tbl.insert().values(val='abc'))`. Right?

PR #69, opened 2015-08-18
Fix missed table create in a SQLAlchemy example

hyzhak:
fix for https://github.com/aio-libs/aiopg/issues/68 issue


asvetlov:
Thanks

PR #72, opened 2015-09-21
Specify test directory when calling py.test

luhn:
I made a virtualenv in the "env" directory inside the project root.  Py.test picked up all the tests from the packages in the virtualenv.  This prevents that from happening.


luhn:
Tests are failing, it looks because cursor timeout took a few milliseconds too long to be raised.

asvetlov:
Thanks

PR #74, opened 2015-09-21
Allow for custom aiopg connection factories

luhn:
This is so I can do this:  https://gist.github.com/luhn/e4972184e1e1e724a430

### DIFF
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -40,7 +40,8 @@ def _enable_hstore(conn):
 
 @asyncio.coroutine
 def connect(dsn=None, *, timeout=TIMEOUT, loop=None,
-            enable_json=True, enable_hstore=True, echo=False, **kwargs):
+            enable_json=True, enable_hstore=True, echo=False,
+            aio_connection_factory=None, **kwargs):
     """A factory for connecting to PostgreSQL.
 
     The coroutine accepts all parameters that psycopg2.connect() does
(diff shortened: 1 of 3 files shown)


luhn:
Tests failing for same reason as #72

asvetlov:
What's your use case for the PR?
`aiopg.Connection` is not intended for inheritance.

luhn:
See the gist I linked above.
On Wed, Sep 23, 2015 at 7:42 AM Andrew Svetlov <notifications@github.com>
wrote:

> What's your use case for the PR?
> aiopg.Connection is not intended for inheritance.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/aio-libs/aiopg/pull/74#issuecomment-142622955>.
>
-- 
Regards,
Theron Luhn
CTO, Guest Innovations

asvetlov:
For your case you shouldn't derive from Connection but embed connection instance inside your class with `.query` and `operation` methods -- like I did it for SQLAlchemy support (https://github.com/aio-libs/aiopg/blob/master/aiopg/sa/connection.py)

luhn:
I feel it's excessive to add another layer of abstraction just for a couple of convenience functions.  But perhaps you're right.  I can't find an example of anybody else subclassing a database connection.

asvetlov:
Let's close the issue.

PR #75, opened 2015-09-21
asynchronous context manager for cursors (experimental)

luhn:
This is to support the following paradigms:

```python
with (yield from conn.cursor()) as cur:
    cur.execute('SELECT 1')
```

And, in Python 3.5:

```python
    cur.execute('SELECT 1')
```

Right now, conn.cursor() returns an awaitable, coroutine-like object, but it is not actually a coroutine.  Not sure if this is a problem.  Python doesn't enforce it.

Potential additional features:

* Implement same functionality for `Pool.cursor()`
* Context manager for transactions, e.g. `async with conn:  # This begins and commits/rollbacks transaction`

I have to step off of this for the moment, so thought I show what I've done so far.

### DIFF
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -40,7 +40,8 @@ def _enable_hstore(conn):
 
 @asyncio.coroutine
 def connect(dsn=None, *, timeout=TIMEOUT, loop=None,
-            enable_json=True, enable_hstore=True, echo=False, **kwargs):
+            enable_json=True, enable_hstore=True, echo=False,
+            aio_connection_factory=None, **kwargs):
     """A factory for connecting to PostgreSQL.
 
     The coroutine accepts all parameters that psycopg2.connect() does
@@ -52,8 +53,13 @@ def connect(dsn=None, *, timeout=TIMEOUT, loop=None,
     if loop is None:
         loop = asyncio.get_event_loop()
 
+    if aio_connection_factory is None:
+        aio_connection_factory = Connection
+
     waiter = asyncio.Future(loop=loop)
-    conn = Connection(dsn, loop, timeout, waiter, bool(echo), **kwargs)
+    conn = aio_connection_factory(
+        dsn, loop, timeout, waiter, bool(echo), **kwargs
+    )
     try:
         yield from conn._poll(waiter, timeout)
     except Exception:
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -434,6 +434,7 @@ class _CursorContextManager:
         self._impl = impl
         self._timeout = timeout
         self._echo = echo
+        self._cursor = None
 
     def __iter__(self):
         impl = yield from self._impl
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -18,6 +18,7 @@ __all__ = ('connect',)
 
 TIMEOUT = 60.0
 PY_341 = sys.version_info >= (3, 4, 1)
+PY_35 = sys.version_info >= (3, 5)
 
 
 @asyncio.coroutine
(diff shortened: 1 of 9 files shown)


aiopg/connection.py
```diff
@@ -190,11 +195,13 @@ def cursor(self, name=None, cursor_factory=None,
...
-                                       scrollable=scrollable,
-                                       withhold=withhold)
-        return Cursor(self, impl, timeout, self._echo)
+        impl = self._cursor(name=name,
+                            cursor_factory=cursor_factory,
+                            scrollable=scrollable,
+                            withhold=withhold)
+        return _CursorContextManager(self, impl, timeout, self._echo)
+
+    cursor._is_coroutine = True
```
asvetlov: This is looks incorrect.

asvetlov:
I believe we should use `async with` in test suite but run these tests on python3.5 only.

See https://github.com/aio-libs/aiorwlock/blob/master/tests/conftest.py#L49-L52 for inspiration

luhn:
Can you specify what looks incorrect about it?

Pytest.mark.skip was my first thought, but "async with" is invalid syntax
in Python <3.5, so it won't even compile.
On Wed, Sep 23, 2015 at 7:46 AM Andrew Svetlov <notifications@github.com>
wrote:

> I believe we should use async with in test suite but run these tests on
> python3.5 only.
>
> See
> https://github.com/aio-libs/aiorwlock/blob/master/tests/conftest.py#L49-L52
> for inspiration
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/aio-libs/aiopg/pull/75#issuecomment-142624864>.
>
-- 
Regards,
Theron Luhn
CTO, Guest Innovations

asvetlov:
1. You shouldn't rely on asyncio implementation details. Either mark method as coroutine by `@asyncio.coroutine` decorator or do nothing.
2. I've added pytest hook to master branch so if you put your test file under `./tests/pep492/test_whatever.py` name it will be ignored on python < 3.5

luhn:
Then it will have to be nothing. Decorating it as a co routine wraps the
function which prevents exposure of __aenter__ and __aexit__.

I apologize, I looked too quickly to realize that's what it did. Neat!
On Wed, Sep 23, 2015 at 8:26 AM Andrew Svetlov <notifications@github.com>
wrote:

>
>    1. You shouldn't rely on asyncio implementation details. Either mark
>    method as coroutine by @asyncio.coroutine decorator or do nothing.
>    2. I've added pytest hook to master branch so if you put your test
>    file under ./tests/pep492/test_whatever.py name it will be ignored on
>    python < 3.5
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/aio-libs/aiopg/pull/75#issuecomment-142639196>.
>
-- 
Regards,
Theron Luhn
CTO, Guest Innovations

aiopg/connection.py
```diff
@@ -421,3 +428,24 @@ def __del__(self):
...
         """Return notification queue."""
         return self._notifies
+
+
+class _CursorContextManager:
+    def __init__(self, conn, impl, timeout, echo):
+        self._conn = conn
+        self._impl = impl
+        self._timeout = timeout
+        self._echo = echo
```
asvetlov: Please add `self._cursor = None` line after this one.

aiopg/connection.py
```diff
@@ -421,3 +428,24 @@ def __del__(self):
...
+        return Cursor(self._conn, impl, self._timeout, self._echo)
+
+    @asyncio.coroutine
+    def __aenter__(self):
+        self._cursor = yield from self
+        return self._cursor
+
+    @asyncio.coroutine
+    def __aexit__(self, type, value, traceback):
+        self._cursor.close()
```
asvetlov: Technically we should check for `self._cursor` existence I believe.

Also `cursor.close()` is buggy now but it is another issue.

asvetlov:
So please rewrite `async with` based tests in new manner.

luhn:
Will do.

aiopg/connection.py
luhn: I don't think that's necessary.  For __aexit__ to be called by the with statement, __aenter__ must have been successfully executed, so the cursor should exist.

luhn:
Done.  pyflakes was complaining about a syntax error, but doesn't support an "--exclude" command line option like pep8 does, so I changed over to flake8 instead.  It uses pyflakes and pep8, just combines them into one interface.

luhn:
Interesting.  Coverage is throwing an exception in 3.5 when testing this branch.  Works fine when testing master, though.  Will investigate further when I get the opportunity.

shongololo:
looking forward to the async with syntax.

luhn:
An issue in coverage is causing the HTML coverage report to fail due to something in `tests/test_pep492.py`.  I've disabled coverage on that file for now.

tests/test_connection.py
```diff
@@ -230,6 +231,16 @@ def go():
 
         self.loop.run_until_complete(go())
 
+    @pytest.mark.xfail
```
asvetlov: Test should be passed

aiopg/connection.py
```diff
@@ -421,3 +420,27 @@ def __del__(self):
     def notifies(self):
         """Return notification queue."""
         return self._notifies
+
+
+class _CursorContextManager:
```
asvetlov: See https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/client.py#L410-L467 for proper context manager implementation

luhn: That's some really interesting code.  Thanks!

To make sure we're clear, what's the goal here?  In my test below, I assert `asyncio.iscoroutine(conn.cursor())`; is this the goal?  What about `asyncio.iscoroutinefunction(conn.cursor))`?  (As far as I can tell, aiohttp does not achieve this.)  Are there any other requirements?

aiohttp [modifies asyncio.coroutines._COROUTINE_TYPES](https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/client.py#L465), which is relying on the structure of CPython implementation.  You chastised me for doing similar previously, what's the difference?  Why is `coroutines._COROUTINE_TYPES += (_RequestContextManager,)` okay, but `cursor._is_coroutine = True` bad?  (Just curious, I'm not offended at having to remove `cursor._is_coroutine`.  It was a last resort to get `asyncio.iscoroutinefunction` to behave.)

asvetlov: To make `loop.run_until_complete(conn.cursor())` works.
`iscoroutine` should pass also I believe.

Implementing `.send` etc. is required by `collections.abc.coroutine`.

`coroutines._COROUTINE_TYPES += (_RequestContextManager,)` hack is required for Python 3.4 for `run_until_complete()`, actually for `ensure_future`.
Guido approved the hack in python/asyncio#277

Requirements are relaxed by `asyncio` master a bit but released version still need that.

luhn:
Hey @asvetlov, anything else that needs to be changed for this to get merged?

asvetlov:
Supersseded by #88

PR #79, opened 2015-10-13
English edit and some doc improvements.

bitdancer:
In addition to improving the English phrasing, this adds instructions
for using venv to set up the virtual environment and a mention
of the setup_test_db.sql script.  I also added a note to the README
that CONTRIBUTING explains how to set up to run the tests.

I don't know if you'll want the English edits; you use "I" in the README
notes, and applying them means it will no longer be in your "normal
voice".


asvetlov:
Thanks!

PR #82, opened 2015-10-17
Create/drop tables using metadata in aiopg.sa

rrader:
Instead of using workarounds like 
```
res = yield from conn.execute(CreateTable(tbl))
```
with this patch it's available to create/drop tables via native sqlalchemy way via metadata
```
meta = sa.AsyncMetaData()
tbl = Table('sa_tbl', meta,
            Column('id', Integer, nullable=False,
                   primary_key=True),
            Column('name', String(255)))
conn = yield from self.connect()
yield from meta.create_all(conn)
```

Not the best implementation, because contains lot of repeated code from SA, but at least for end user it's more transparent.

### DIFF
--- a/aiopg/sa/__init__.py
+++ b/aiopg/sa/__init__.py
@@ -3,6 +3,7 @@ from .connection import SAConnection
 from .engine import create_engine, Engine
 from .exc import (Error, ArgumentError, InvalidRequestError,
                   NoSuchColumnError, ResourceClosedError)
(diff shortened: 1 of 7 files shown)


asvetlov:
I cannot accept the PR.
1. It borrows too much code from SQLAlchemy.
2. The code is not 100% test covered.

rrader:
yep, it's just an experiment.

PR #84, opened 2015-10-30
If execute times out, handle possible cancel timeout.

bitdancer:
execute was fixed to close the cursor if it timed out, to prevent a subsequent 'query in flight' error.  However, the test assumes that cancel will not time out.  This patch introduces a test for the case where the cancel *also* times out, and fixes it by ignoring the cancel timeout in the execute timeout error recovery.

I assume that extending the original fix in this way is correct, but I'm not really sure about what happens to the transaction that timed out from aiopg's point of view.  In my application code I'm currently dealing with the timeout by checking the DB before doing a retry...I'll find out some time in the next couple of days if that actually works :)

### DIFF
--- a/aiopg/cursor.py
+++ b/aiopg/cursor.py
@@ -111,7 +111,10 @@ class Cursor:
         try:
             yield from self._conn._poll(waiter, timeout)
         except asyncio.TimeoutError:
-            yield from self._conn.cancel()
+            try:
+                yield from self._conn.cancel()
+            except asyncio.TimeoutError:
+                pass
             self._impl.close()
             raise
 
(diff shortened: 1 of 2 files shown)


bitdancer:
Looks like there is a timing issue with the test, I presume because of the cancel transaction that we don't actually care about).  I'll rewrite the test to mock the cancel.

bitdancer:
After further investigation, I don't think my fix is valid.  The test only passes, I think, because the cancel is executed.  If I mock out Connection.cancel, we are back to the ProgrammingError.

It looks like this needs a fix upstream in psycopg2.

asvetlov:
I believe it's fixed by 2a72b6097e9d0a4c23318b03e57e10767acbbdd4

PR #88, opened 2015-12-19
Add async with support for Pool, Connection, Cursor

jettify:
1) Added support for async context managers for `Pool`, `Connection`, `Cursor`, `connect`, `create_pool`
2) Added async iteration support for cursor

I have not added context manager for `pool.cursor` since `aiomysql` and `aioodbc` do not support this API.


asvetlov:
I believe travis doc about caching is not correct, at least I can see caching setting for aiohttp project.

asvetlov:
please fix travis config file

jettify:
If I bring back that config, travis will not start build due to .travisci validation error, see: https://travis-ci.org/aio-libs/aiopg/builds/97873859

asvetlov:
Ok, just fix travis file.
I don't care too much about caching.

aiohttp use the following lines:
```
cache:
  directories:
  - $HOME/.cache/pip
before_cache:
  - rm -f $HOME/.cache/pip/log/debug.log
```

so it's not apt cache.

jettify:
Do not follow, you mean fix .travisci in separate PR? Because right now travis config same as you posted.

jettify:
I have used some code and tricks from this PR https://github.com/aio-libs/aiopg/pull/75, it will be obsolete once this PR merged...

PR #90, opened 2016-01-04
Add async with for engine

jettify:
This PR adds `async with` support for `Engine`:
```python
async with create_engine(...) as engine:
    pass
```
as well as:
```python
engine = await create_engine(...)
    conn.execute()
```


lerrua:
+1

asvetlov:
thanks

Issue #92, opened 2016-01-05
Support async iterator in ResultProxy


asvetlov:
```
async for row in conn.exectute("SELECT ..."):
    print(row)
```

jettify:
I tried to implement and found that this API is not very good, because there is no way to close created `ResultProxy` object with cursor inside.  We should encourage:
```python
async with conn.execute("SELECT ...") as result:
    async for record in result:
        print(record)
```

asvetlov:
Maybe. I used dirty trick to close result IIRC.

jettify:
resolved by https://github.com/aio-libs/aiopg/pull/94

Issue #91, opened 2016-01-05
Support async conext manager for transaction


asvetlov:
```async with conn.begin():
       conn.execute(...)
```

jettify:
resolved by #94

PR #94, opened 2016-01-05
Add more async support

jettify:
PR for #92 #91

### DIFF
--- a/aiopg/sa/result.py
+++ b/aiopg/sa/result.py
@@ -5,7 +5,16 @@ from collections.abc import Mapping, Sequence
 
 from sqlalchemy.sql import expression, sqltypes
 
+from ..utils import PY_35
 from . import exc
+
+
+try:
+    StopAsyncIteration
+except NameError:
+    class StopAsyncIteration(Exception):
+        """Just stab for StopAsyncIteration from python 3.5"""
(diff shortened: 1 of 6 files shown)


jettify:
Fixed issue with possible double commit. Please review.

aiopg/sa/result.py
```diff
@@ -5,9 +5,18 @@
 
 from sqlalchemy.sql import expression, sqltypes
 
+from ..utils import PY_35
 from . import exc
 
 
+try:
+    StopAsyncIteration
```
asvetlov: Why do you need the check?

`if PY_35:` statement above prevents `StopAsyncIteration` usage for python < 3.5

asvetlov:
Looks good but probably we should get rid of `StopAsyncIteration` in `utils.py` too.

jettify:
I have dropped  `StopAsyncIteration` but pyflakes complains with:
```python
aiopg/cursor.py:392:23: F821 undefined name 'StopAsyncIteration'
aiopg/sa/result.py:337:23: F821 undefined name 'StopAsyncIteration'
```
so I have added `# noqa` marker to those lines.

asvetlov:
LGTM

PR #103, opened 2016-01-20
Enable uuid feature

hellysmile:


tests/test_extended_types.py
```diff
@@ -25,7 +26,7 @@ def connect(self, **kwargs):
         cur = yield from conn.cursor()
         yield from cur.execute("DROP TABLE IF EXISTS tbl")
         yield from cur.execute("""CREATE TABLE tbl (
-                                      id SERIAL,
+                                      id UUID,
```
jettify: May be better to have separate test for UUID extension?

jettify:
Looks good but I have couple questions:
1) do we need enable uuid by default? 
2) Is uuid feature available out of box? if not what happened in case of postgres extension package is not installed?

hellysmile:
1) There is by default JSON and HSTORE, why do not have UUID?
2) UUID is built-in PostgreSQL data type, as I remember since 8.?

jettify:
Looks good to me, @asvetlov any objections to merge this PR?

GMLudo:
To my experience, uuid isn't built-in: at least with postgresql 9.4 on Debian, you need to install postgresql-contrib and enable extension uuid-ossp.

hellysmile:
uuid-ossp is for generation UUID via PostgreSQL server.
UUID data type is not uuid-ossp

GMLudo:
@hellysmile thanks for the info.
+1 for that then.

asvetlov:
Perfect!

PR #107, opened 2016-03-04
Dockerize pg server

asvetlov:


asvetlov:
@popravich @Drizzt1991 you, guys, may be curious about my vision of tests dockerization.

I believe using the similar approach in aioredis and aiokafka makes sense.

popravich:
This looks good for travis but to run tests locally I (anyone) would need to have +1 dependency - docker - on their system

asvetlov:
Really it's +1/-1 dependency: you exchange Postgres server with Docker daemon.

PR #114, opened 2016-04-15
Fixed doc mistakes in index.srt

Morreski:


docs/index.rst
```diff
@@ -48,11 +48,11 @@ See example::
 
     async def go():
         pool = await aiopg.create_pool(dsn)
-        async with pool.acquire() as cur:
+        async with pool.acquire() as conn:
```
asvetlov: Good catch

docs/index.rst
```diff
@@ -48,11 +48,11 @@ See example::
...
             async with conn.cursor() as cur:
                 await cur.execute("SELECT 1")
                 ret = []
-                async for row in cur:
+                for row in await cur.fetchall():
```
asvetlov: Async iteration over executed cursor is allowed as shown in original line.

Morreski: You're right, i didn't catch the difference with this __aiter__ method that is called during async iteration. I think i get it now...

PR #119, opened 2016-06-02
asyncio.async() is deprecated in favor of asyncio.ensure_future()

rudyryk:
Not sure if we really need to support an older version, but I think that backwards compatibility is generally a good idea.


jettify:
Good point, but lets create `create_future` helper in utils like aiohttp does

rudyryk:
@jettify Ah, thanks for the tip, I'll take a look :)

PR #124, opened 2016-07-14
Don't crash when Connection is GC'ed

### LINKED ISSUE #111: aiopg closes event loop on failed connection
When a connection to postgres fails, the asyncio event loop is closed. This prevents me from running any further coroutines (in particular, even those unrelated to aiopg) after a connection fails, which isn't always desirable.

mpaolini:
`Connection` `_conn` attribute might not be set if `__init__` crashes early
due to a psycopg error on connect.

Make sure `__del__` does not crash when this happens.

Closes #111

### DIFF
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -449,7 +449,11 @@ class Connection:
 
     if PY_341:  # pragma: no branch
         def __del__(self):
-            if not self._conn.closed:
+            try:
+                _conn = self._conn
+            except AttributeError:
+                return
+            if _conn is not None and not _conn.closed:
                 self.close()
                 warnings.warn("Unclosed connection {!r}".format(self),
                               ResourceWarning)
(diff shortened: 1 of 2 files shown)


aiopg/connection.py
```diff
@@ -449,7 +449,11 @@ def echo(self):
 
     if PY_341:  # pragma: no branch
         def __del__(self):
-            if not self._conn.closed:
+            try:
+                _conn = self._conn
+            except AttributeError:
+                return
+            if _conn is not None and not _conn.closed:
```
jettify: May be this check should be moved inside `self.close()`, so finalizing method will be idempotent?

mpaolini: I am not sure it is a good idea: the whole point of this `__del__` method is issuing a warning and calling `self.close()` if there was an issue. Making `close()` resitent to `self._conn` not being set will clutter that method for no reson IMHO

asvetlov:
Thanks

Issue #16, opened 2014-06-16
Add timeout for acquiring from pool


asvetlov:

serg666:
Hi! Any news on this issue?

mpaolini:
I think this can be closed. see comments in #125

asvetlov:
Use `asyncio.wait_for` or `aiohttp.Timeout`.

PR #125, opened 2016-07-15
Add timeout acquiring from pool

mpaolini:
Stab at issue #16 

TODO:

- [ ] tests

### DIFF
--- a/aiopg/pool.py
+++ b/aiopg/pool.py
@@ -153,13 +153,13 @@ class Pool(asyncio.AbstractServer):
 
         self._closed = True
 
-    def acquire(self):
+    def acquire(self, timeout=None):
         """Acquire free connection from the pool."""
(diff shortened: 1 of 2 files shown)


jettify:
asyncio usually do not path timeout as kw argument in favour `wait` and `wait_for`, I think we should follow aiohttp:
```python
with aiohttp.Timeout(0.001):
    async with aiohttp.get('https://github.com') as r:
        await r.text()
```
this context manager is present in new asyncio and packaged with aiohttp.

mpaolini:
so you are suggesting we close #16 ?

asvetlov:
Unfortunately `asyncio.timeout` was ripped out from asyncio due problems with tornado.
`aiohttp.Timeout` is working and will present for very long time.

So yes, I suggest using timeout context manager and don't add timeout parameter.

PR #139, opened 2016-09-05
Remove callbacks from a bad file descriptor immediately

### LINKED ISSUE #138: Connection error occasionally affects another connection
Symptoms:
When creating a bunch of aiopg connections in parallel, if one of them fails with a psycopg error, **another** one might fail with tracebacks like these:

```
  File "/home/lena/.virtualenvs/tsdebug/lib/python3.5/site-packages/aiopg/connection.py", line 213, in _poll
    yield from asyncio.wait_for(self._waiter, timeout, loop=self._loop)
  File "/usr/local/lib/python3.5/asyncio/tasks.py", line 396, in wait_for
    raise futures.TimeoutError()
concurrent.futures._base.TimeoutError
```

```
ERROR:api.query:Traceback (most recent call last):
  File "/usr/local/lib/python3.5/asyncio/tasks.py", line 241, in _step
    result = coro.throw(exc)
  File "/home/lena/xxx/xxx/xxx/xxx/xxx/api/postgresql_querying.py", line 140, in query_db
    cursor_factory=NamedTupleCursor) as conn:
  File "/home/lena/.virtualenvs/tsdebug/lib/python3.5/site-packages/aiopg/utils.py", line 75, in __aenter__
    self._obj = yield from self._coro
  File "/home/lena/.virtualenvs/tsdebug/lib/python3.5/site-packages/aiopg/connection.py", line 69, in _connect
    yield from conn._poll(waiter, timeout)
  File "/home/lena/.virtualenvs/tsdebug/lib/python3.5/site-packages/aiopg/connection.py", line 216, in _poll
    yield from asyncio.shield(cancel(), loop=self._loop)
  File "/usr/local/lib/python3.5/asyncio/futures.py", line 361, in __iter__
    yield self  # This tells Task to wait for completion.
  File "/usr/local/lib/python3.5/asyncio/tasks.py", line 296, in _wakeup
    future.result()
  File "/usr/local/lib/python3.5/asyncio/futures.py", line 274, in result
    raise self._exception
  File "/usr/local/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
  File "/home/lena/.virtualenvs/tsdebug/lib/python3.5/site-packages/aiopg/connection.py", line 206, in cancel
    self._conn.cancel()
psycopg2.OperationalError: asynchronous connection attempt underway
```

Reason:
Python3.5 only allows to add one reader to a file descriptor, and add_reader method returns None. Thus remove_reader method removes whatever reader is currently registered on the descriptor.
So the following happens sometimes (once in 10–25 requests in my experience):

1. aiopg's Connection() object initializes, adds a reader on a file descriptor
2. this connection object gets a psycopg2.Error when polling
3. another aiopg Connection() object is initialized, it adds a reader on the same file descriptor (which effectively cancels the reader added in step 1)
4. the first connection object's .close() is called, and removes the reader from the file descriptor

As the reader removed actually "belonged" to the second connection (and not to the first, which removed it), this creates problems for it.

Suggested solution (thanks @asvetlov):
If a psycopg2.Error occurs, **and** the file descriptor has gone bad, remove the reader immediately (don't wait for .close() to do it)

not-even:
Previously, the callback removal was postponed until .close() call, and
sometimes this .close() call cleared a file descriptor from a callback set
by another connection object (if the file desriptor was closed by psycopg2 when
an error occurred, and then opened again for a new connection).

The effect of this was, when making tens of parallel connections, an error in
one of them made another time out.

Now, if a psycopg2 error occurs, and the file descriptor has gone bad, the
callback is removed immediately, and the aiopg connection object 'forgets'
about this file descriptor (self._fileno = None), and doesn't try to remove
callbacks in .close().

If there were no errors, callbacks are still removed during .close() call.

Fixes #138.

### DIFF
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -1,4 +1,5 @@
 import asyncio
+import contextlib
 import errno
 import fcntl
 import sys
@@ -126,11 +127,8 @@ class Connection:
                     fcntl.fcntl(self._fileno, fcntl.F_GETFD)
                 except OSError as os_exc:
                     if os_exc.errno == errno.EBADF:
-                        try:
+                        with contextlib.suppress(OSError):
                             self._loop.remove_reader(self._fileno)
-                        except OSError as os_exc2:
-                            if os_exc2.errno == errno.EBADF:
-                                pass
                 finally:
                     # forget a bad file descriptor, don't try to touch it
                     self._fileno = None
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -1,5 +1,6 @@
 import asyncio
 import errno
+import fcntl
 import sys
 import traceback
 import warnings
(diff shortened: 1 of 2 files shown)


aiopg/connection.py
```diff
@@ -120,10 +121,25 @@ def _ready(weak_self):
                 notify = self._conn.notifies.pop(0)
                 self._notifies.put_nowait(notify)
         except (psycopg2.Warning, psycopg2.Error) as exc:
+            if self._fileno is not None:
+                try:
+                    fcntl.fcntl(self._fileno, fcntl.F_GETFD)
+                except OSError as os_exc:
+                    if os_exc.errno == errno.EBADF:
+                        try:
```
asvetlov: `with contextlib.suppress(OSError)` should be better than another `try/except`.

not-even: Agreed. Updated

asvetlov:
Thanks!

PR #142, opened 2016-09-17
Fixed connection to work under both windows and posix based systems

ryanhodge:
Also updated tests to always remove descriptors from event loop if closed. This is because the asyncio event loop under windows creates errors if the loop is run and there are closed file descriptors in the select. 

The reason that aiopg stopped working under windows (it did work until v11.0) was the recent introduction of the use of the fcntl function to determine if the filedescriptor had been closed. A more OS friendly poll on the socket has been used instead, to determine if the filedescriptor has been closed.

To run the tests under windows, the `pg_server` fixture had to be changed, as obviously docker is not available under windows. It was changed to run on a local database rather than a docker database, but these changes were not added to the pull request as they were a hack, and maybe not the way that developers would want to run the tests (they exist in a localtest branch on my fork).

### DIFF
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -1,11 +1,12 @@
 import asyncio
 import contextlib
 import errno
-import fcntl
+import select
 import sys
 import traceback
 import warnings
 import weakref
+import platform
 
 import psycopg2
 from psycopg2.extensions import (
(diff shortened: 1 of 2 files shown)


asvetlov:
Please fix flake8 blames.
Let's keep `pg_server` as is for now at least.
We could change it to support both dockerized and undockerized configurations but I don't want to carry extra maintenance burden. This bug is the first OS-specific problem during all aiopg existence.

asvetlov:
Thanks

Issue #189, opened 2016-10-21
hstore with cursor_factory=psycopg2.extras.RealDictCursor Errors


bgcgarlipp:
Hi

When intializing a new connection with cursor_factory=psycopg2.extras.RealDictCursor on a databse with hstore extension on, you get an error as follows. This is due to a dict being returned as result and not list.

> Traceback (most recent call last):
  File "wsclient.py", line 134, in <module>
    loop.run_until_complete(db.init_db()) 
  File "/usr/lib/python3.5/asyncio/base_events.py", line 337, in run_until_complete
    return future.result()
  File "/usr/lib/python3.5/asyncio/futures.py", line 274, in result
    raise self._exception
  File "/usr/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
  File "/home/bernhardt/git_work/qs_switch/QS_Workflow/client/async_quantdb.py", line 25, in init_db
    cursor_factory=psycopg2.extras.RealDictCursor)
  File "/home/bernhardt/git_work/qs_switch/QS_Workflow/pyenv/lib/python3.5/site-packages/aiopg/utils.py", line 57, in __await__
    resp = yield from self._coro
  File "/home/bernhardt/git_work/qs_switch/QS_Workflow/pyenv/lib/python3.5/site-packages/aiopg/pool.py", line 45, in _create_pool
    yield from pool._fill_free_pool(False)
  File "/home/bernhardt/git_work/qs_switch/QS_Workflow/pyenv/lib/python3.5/site-packages/aiopg/pool.py", line 188, in _fill_free_pool
    **self._conn_kwargs)
  File "/home/bernhardt/git_work/qs_switch/QS_Workflow/pyenv/lib/python3.5/site-packages/aiopg/utils.py", line 52, in __iter__
    resp = yield from self._coro
  File "/home/bernhardt/git_work/qs_switch/QS_Workflow/pyenv/lib/python3.5/site-packages/aiopg/connection.py", line 78, in _connect
    oids = yield from _enable_hstore(conn)
  File "/home/bernhardt/git_work/qs_switch/QS_Workflow/pyenv/lib/python3.5/site-packages/aiopg/connection.py", line 37, in _enable_hstore
    rv0.append(oids[0])
KeyError: 0
> 

How to reproduce:
install 

> sudo apt-get install postgresql-contrib
create hstore extension on database

> CREATE EXTENSION hstore; 

[...]

PR #190, opened 2016-10-22
Fix for Issue 189, resolved hstore wiht realDictCursor error

bgcgarlipp:
This is my proposed solution for issue #189.
It sesolves the problem where the query result is returned as a dict and not a list due to the cursor_factory being a RelDictRow

### DIFF
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -39,8 +39,12 @@ def _enable_hstore(conn):
         """)
     rv0, rv1 = [], []
     for oids in (yield from cur.fetchall()):
-        rv0.append(oids[0])
-        rv1.append(oids[1])
+        if type(oids) is list:
+            rv0.append(oids[0])
+            rv1.append(oids[1])
+        else:
+            rv0.append(oids['oid'])
+            rv1.append(oids['typarray'])
 
     cur.close()
     return tuple(rv0), tuple(rv1)


bgcgarlipp:
I see it fails on this test however I am not sure why
 tests/test_connection.pytest_cancel_pending_op[9.5] ⨯

aiopg/connection.py
```diff
@@ -39,8 +39,12 @@ def _enable_hstore(conn):
         """)
     rv0, rv1 = [], []
     for oids in (yield from cur.fetchall()):
-        rv0.append(oids[0])
-        rv1.append(oids[1])
+        if type(oids) is list:
```
jettify: `isinstance()` should be more robust than direct type comparisons.

jettify:
Could you please add simple test, like create new connection with `RealDictCursor` and `enable_hstore=True` and execute simple query `select 1` or something ? This will help us not to break this feature accidentally in featue.

aiopg/connection.py
bgcgarlipp: I'll do that

bgcgarlipp:
I'll add a test later today or tomorrow.

asvetlov:
ping

jettify:
@bgcgarlipp sorry for bothering, do you have time to finish PR?

Pliner:
Fixed by https://github.com/aio-libs/aiopg/commit/57b5d55b3593ed6c6d8507d33a8bf1e0300fce84

PR #204, opened 2016-11-14
Tests cleanup

jettify:
* removed unused tox
* removed flag USE_SA=0
* test 3.5-dev changed to 3.6-dev


jettify:
@asvetlov please review

asvetlov:
Thanks!

asvetlov:
There is another issue with running tests on travis: too big matrix takes too long for finishing.
I've changed aiohttp travis config to use the only run per python version but execute test code with different environment variables like `PYTHONAYNCIODEBUG=x` etc.

Maybe we need to do the same for all aio-libs.

Issue #49, opened 2015-03-06
Default value doesn't work


AnderUstarroz:
I've found an unexpected behavior in SQLAlchemy. I'm using the following versions:
- **SQLAlchemy** (0.9.8)
- **PostgreSQL** (9.3.5)
- **psycopg2** (2.5.4)
- **aiopg** (0.5.1)

This is the table definition for the example:

```
import asyncio
from aiopg.sa import create_engine
from sqlalchemy import (
  MetaData,
  Column,
  Integer,
  Table,
  String,
)
metadata = MetaData()

users = Table('users', metadata,
  Column('id_user', Integer, primary_key=True, nullable=False),
  Column('name', String(20), unique=True),
  Column('age', Integer, nullable=False, default=0),
)
```

Now if I try to execute a simple insert to the table just populating the **_id_user**_ and **_name**_, the column **_age**_ should be auto-generated with the default value **0** right? Lets see...

```
@asyncio.coroutine
def go():
  engine = yield from create_engine('postgresql://USER@localhost/DB')
  data = {'id_user':1, 'name':'Jimmy' }
  stmt = users.insert(values=data, inline=False)
  with (yield from engine) as conn:
    result = yield from conn.execute(stmt)


loop = asyncio.get_event_loop()
loop.run_until_complete(go())
```

This is the resulting statement with the corresponding error:

```
INSERT INTO users (id_user, name, age) VALUES (1, 'Jimmy', null);

psycopg2.IntegrityError: null value in column "age" violates not-null constraint
```

I didn't provide the **age** column, so where is that **age = null** value coming from? I was expecting something like this:

```
INSERT INTO users (id_user, name) VALUES (1, 'Jimmy');
```

Or if the **_default**_ flag actually works should be:

```
INSERT INTO users (id_user, name, Age) VALUES (1, 'Jimmy', 0);
```

Could you put some light on this?

asvetlov:
Yes, you are right.
aiopg now ignores 'default' argument on data manipulations.
You can use server defaults:

```
        yield from conn.execute('''CREATE TABLE users (
                                            id_user serial PRIMARY KEY,
                                            name varchar(20),
                                            age integer DEFAULT 0)''')
```
or 
```

from sqlalchemy.schema import CreateTable

users = Table('users', metadata,
              Column('id_user', Integer, primary_key=True, nullable=False),
              Column('name', String(20), unique=True),
              Column('age', Integer, nullable=False, server_default='0'))

[...]

PR #206, opened 2016-11-15
sa default # 49

vir-mir:
Hi!
I offer a patch for the default fields.

Do you think it makes sense for life?

@asvetlov 

Issues https://github.com/aio-libs/aiopg/issues/49


asvetlov:
Wow! Very impressive.
Let me sleep on it and make careful review.

vir-mir:
- Combed code.
- Add processing scalar, sequence and fix set value None where default not null.

vir-mir:
@asvetlov any news?

pomidoroshev:
@asvetlov Чё как?

Eduard90:
We need default values!

aiopg/sa/engine.py
```diff
@@ -2,20 +2,53 @@
...
+        for key, default in default_fields:
+            pd[key] = (yield from self._exec_default(cursor, default))
+
+        return pd
+
+    @asyncio.coroutine
+    def _exec_default(self, cursor, default):
+        if default.is_sequence:
+            sql = "select nextval('{}')".format(
+                self.dialect.identifier_preparer.format_sequence(default))
```
vir-mir: sqlalchemy makes a queries to the db.
Of course, we can make queries in advance...
such behavior we should document it explicitly

jettify:
@vir-mir thank you very much for this PR! Lets give @asvetlov chance to review on weekends, otherwise I will merge it on Monday.

asvetlov:
Looks good.
@jettify please merge

serg666:
Guys, what about this [comment](https://github.com/aio-libs/aiopg/issues/49#issuecomment-263150311) ?

Eduard90:
I have problems with UUID (PostgreSQL) column type:
```
  File "/usr/lib/python3.5/uuid.py", line 137, in __init__
    hex = hex.replace('urn:', '').replace('uuid:', '')
AttributeError: 'UUID' object has no attribute 'replace'
```

Table:
```
InvoiceTable = NewTable('invoice', metadata,
                        Column('id', UUID, primary_key=True),
                        Column('affiliate_id', UUID),
                        Column('status', String, index=True, default=StatusEnum.not_approved.name),
                        Column('created_at', DateTime, default=datetime.now),
                        Column('deleted', Boolean, default=False),
                        Column('is_paid', Boolean, default=False),
                        )
```

Primary key (id) is UUID and "by default" it must generate new UUID on insert.
I have "monkey patch" for _execute method (class SAConnection) but it's very bad code ))

vir-mir:
@Eduard90 

http://docs.sqlalchemy.org/en/latest/core/custom_types.html#backend-agnostic-guid-type
```python
class UUID(sa.types.TypeDecorator):
    impl = sa.String

    def __init__(self, length=255, *args, **kwargs):
        super().__init__(length=255, *args, **kwargs)

    def process_bind_param(self, value, dialect=None):
        if value and isinstance(value, uuid.UUID):
            return value.bytes
        elif value and not isinstance(value, uuid.UUID):
            raise ValueError('value %s is not a valid uuid.UUID' % value)
        else:
            return None

    def process_result_value(self, value, dialect=None):
        if value:
            return uuid.UUID(bytes=value)
        else:
            return None

    def is_mutable(self):
        return False

meta = sa.MetaData()
tbl = sa.Table('sa_tbl4', meta,
               sa.Column('id', UUID(), primary_key=True, default=uuid.uuid4))

...

yield from conn.execute(tbl.insert().values())
yield from conn.execute(tbl.insert().values())
yield from conn.execute(tbl.insert().values())
```

jettify:
@vir-mir thanks! should we put this snippet to the FAQ?

vir-mir:
@serg666 https://github.com/aio-libs/aiopg/pull/56#issuecomment-95560109

vir-mir:
@jettify all say that it is necessary to use an approach TypeDecorator

Eduard90:
@vir-mir Yes, i know about this way. But ... with this custom type Alembic generate not correct migrations.
This problem in Alembic maybe. But i want use UUID and migrations now )))

And sqlalchemy has "builtin" type:
`from sqlalchemy.dialects.postgresql import UUID`
I think it's more correct than custom type. Or not?

Eduard90:
@vir-mir How i can talk with you directly? Skype, Telegram, etc?

vir-mir:
This is problem alembic
```bash
  File "../site-packages/sqlalchemy/sql/schema.py", line 79, in _init_items
    item._set_parent_with_dispatch(self)
AttributeError: 'UUID' object has no attribute '_set_parent_with_dispatch'
```
you can use
`from sqlalchemy.dialects.postgresql import UUID`

But to make it work you need to add a parameter `default`

 ```python
from sqlalchemy.dialects.postgresql import UUID
import uuid

meta = sa.MetaData()
tbl = sa.Table('sa_tbl4', meta,
               sa.Column('id', UUID(as_uuid=True), primary_key=True, default=uuid.uuid4))

...

yield from conn.execute(tbl.insert().values())
yield from conn.execute(tbl.insert().values())
yield from conn.execute(tbl.insert().values())
```

My skype `virmir49`

Issue #86, opened 2015-11-08
Cannot make db -> python value conversion working with custom SA columns


avanov:
The definition:

``` python
import logging
import sqlalchemy.types as types
from enum import Enum as PythonEnum

log = logging.getLogger(__name__)

class PythonMappedEnum(types.TypeDecorator):
    """ Implements mapping between Postgres' Enums and Python Enums.
    """
    impl = types.Enum

    def __init__(self, python_enum_type: PythonEnum, **kwargs):
        self.python_enum_type = python_enum_type
        self.kwargs = kwargs
        enum_args = [x.value for x in python_enum_type]
        super(PythonMappedEnum, self).__init__(*enum_args, **self.kwargs)

    def process_bind_param(self, value: PythonEnum, dialect):
        """ Convert to postgres value
        """
        return value.value

    def process_result_value(self, value: str, dialect):
        """ Convert to python value
        """
        log.debug("=====================")
        log.debug("Called")
        for __, case in self.python_enum_type.__members__.items():
            if case.value == value:
                return case
        raise TypeError("Cannot map Enum value '{}' to Python's {}".format(
            value, self.python_enum_type
        ))

    def copy(self):
        return PythonMappedEnum(self.python_enum_type, **self.kwargs)
```

The calling code (abstract):

``` python
result = yield from SAPoolConnection.execute(SATable.select().limit(1))
data = yield from result.fetchone()
```

When `data` is processed, the value that corresponds to the custom Enum field is of type `str`, because the `process_result_value()` never gets called. But for insert statements, `process_bind_param()` is called as expected.

avanov:
[Relevant docs section](http://docs.sqlalchemy.org/en/latest/core/custom_types.html#augmenting-existing-types)

TimNN:
I was about to report the same issue (took me a while to track this down).

As far as I can tell, the problem is that `aiopg` converts the types solely based on the `oid` of the type reported by postgresql (during my test `process_bind_param` does also not seem to be called I don't know how conversion happens there).

TimNN:
@asvetlov you seem to be the main maintainer of this project, could you maybe chime in on this issue? 

Is this (apart from this issue) a known weakness of aiopg and are there any plans to fix this? Or is there a technical reason that this is (currently) not possible to implement?

asvetlov:
Guys, honestly I did not try custom types yes, so I don't know where it fails.

Maybe I'll find a time for investigation after fixing  #20

[...]

PR #55, opened 2015-04-17
get rid of ResultProxy._result_map

popravich:
Because `SQLCompiler.result_map` API has been removed in SQLAlchemy 1.0
(sqlalchemy diff https://github.com/zzzeek/sqlalchemy/commit/9854114f578833aee49b65e440319a0ec26daab6#diff-6e10aff76d8add452655df8180dc404bL667)

PR #217, opened 2016-11-24
python types columns issues #86

vir-mir:
As I understand it, the problem arose after was merged #55 
effects of problem #86

### DIFF
--- a/aiopg/sa/connection.py
+++ b/aiopg/sa/connection.py
@@ -96,6 +96,9 @@ class SAConnection:
                     processed_parameters.append(params)
                 post_processed_params = self._dialect.execute_sequence_format(
                     processed_parameters)
+
+                # _result_columns is a private API of Compiled,
+                # but I couldn't find any public API exposing this data.
                 result_map = compiled._result_columns
 
             else:
--- a/aiopg/sa/connection.py
+++ b/aiopg/sa/connection.py
@@ -12,7 +12,6 @@ from ..utils import _SAConnectionContextManager, _TransactionContextManager
 
 
 class SAConnection:
-
     def __init__(self, connection, engine):
         self._connection = connection
         self._transaction = None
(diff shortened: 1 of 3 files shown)


jettify:
@popravich could you also take a look?

popravich:
AFAIR, `_result_colums` is remove in sqlalchemy 1.0, so this might not work well with slqalchemy 1+.
Need to take a deeper look

vir-mir:
I think

vir-mir:
I could not find information about the removal _result_colums except that commit https://github.com/zzzeek/sqlalchemy/commit/9854114f578833aee49b65e440319a0ec26daab6

But as I understand it, it affects only DefaultExecutionContext.
I also could not find the use of DefaultExecutionContext in aiopg.
But, for the formation of `result_column_struct` used `compiled.result _result_columns` https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/engine/default.py#L574

I also use `compiled._result_columns` to form column types https://github.com/vir-mir/aiopg/blob/3b4eaa9eb09377fb8c8ecd28e0fa5e14b1aefa95/aiopg/sa/connection.py#L99

Information about the `compiled._result_columns` removal I have not found.

@popravich what do you think?

popravich:
What version of sqlalchemy do you use?

vir-mir:
SQLAlchemy==1.1.4

In travis too SQLAlchemy https://travis-ci.org/aio-libs/aiopg/jobs/178679293#L467

everything is working :slightly_smiling_face:

popravich:
ok,
`_result_columns` is a private API of Compiled, so using it kinda inconvinient,
but I couldn't find any public API exposing this data.
I think we need a comment near `compiled._result_columns` explaining why this private api is used.

vir-mir:
ok, today add.

I just looked, what the normal use of private API https://github.com/aio-libs/aiopg/blob/master/aiopg/sa/result.py#L119

jettify:
@popravich @asvetlov thoughts? LGTM?

asvetlov:
LGTM

Issue #130, opened 2016-07-22
Implement timeout on acquiring connection from pool


asvetlov:

Daenyth:
I believe this is done

PR #218, opened 2016-11-25
Timeout on aquiring connection from pool

beezz:
This PR implements timeout on acquiring connection from pool using the [async-timeout](https://github.com/aio-libs/async-timeout) lib. (#130)

### DIFF
--- a/aiopg/pool.py
+++ b/aiopg/pool.py
@@ -4,6 +4,7 @@ import sys
 import warnings
 
 
+import async_timeout
 from psycopg2.extensions import TRANSACTION_STATUS_IDLE
 
 from .connection import connect, TIMEOUT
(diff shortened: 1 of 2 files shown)


beezz:
I'd love to provide some tests for this as well, but can't think of reasonable way. Any suggestions? Maybe connection to non-existent server with zero timeout would do the trick.

jettify:
Thank you for your contribution, but this change will break library for tornado users, for more information see: https://github.com/KeepSafe/aiohttp/issues/1180

stopspazzing:
Latest info related to timeout and aiohttp here: https://github.com/aio-libs/aiohttp/issues/1524

Should this PR be updated to merge or closed?

asvetlov:
The PR should wait for Tornado 5 I think.

PR #223, opened 2016-11-30
Fix #216

### LINKED ISSUE #216: Random "cursor.execute() called while another coroutine is already waiting for incoming data" under load
I have a trivial `aiohttp.web` application that executes SQL requests via `aiopg` sqlalchemy integration. It's as simple as:

    import aiohttp.web
    from aiopg.sa import create_engine

    app = aiohttp.web.Application()

    async def rows(request):
        async with request.app["db"].acquire() as db:
            return aiohttp.web.json_response(list(await db.execute("SELECT * FROM table")))

    app.router.add_route("GET", "/rows", rows)

    async def init(app):
        app["db"] = await create_engine(host="postgres", user="visio", password="visio", database="visio")

    if __name__ == "__main__":
        handler = app.make_handler()

        loop.run_until_complete(init(app))

        loop.run_until_complete(loop.create_server(handler, "0.0.0.0", 80))
        loop.run_forever()

When server load reaches 100 rps, this error starts appearing randomly:

    RuntimeError: cursor.execute() called while another coroutine is already waiting for incoming data
      File "aiohttp/server.py", line 261, in start
        yield from self.handle_request(message, payload)
      File "aiohttp/web.py", line 88, in handle_request
        resp = yield from handler(request)
      File "visio_longer/views/communicate/__init__.py", line 72, in legacy_communicate
        device = await query_device(db, access_token)
      File "visio_longer/views/communicate/__init__.py", line 31, in query_device
        (Device.access_token == access_token)
      File "aiopg/utils.py", line 72, in __await__
        resp = yield from self._coro
      File "aiopg/sa/connection.py", line 103, in _execute
        yield from cursor.execute(str(compiled), post_processed_params[0])
      File "aiopg/cursor.py", line 103, in execute
        waiter = self._conn._create_waiter('cursor.execute')
      File "aiopg/connection.py", line 186, in _create_waiter
        'already waiting for incoming data' % func_name)

It happens with random queries at random times, once in a few days, sometimes these errors come in bunch of 4 or 2. Is there something wrong with my code? `aiohttp` and `aiopg` versions are latest from pip

themylogin:
Details of what exactly is being fixed are described in the issue comments

### DIFF
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -113,6 +113,8 @@ class Connection:
         self._timeout = timeout
         self._waiter = waiter
         self._writing = False
+        self._cancelling = False
+        self._cancellation_waiter = None
         self._echo = echo
         self._notifies = asyncio.Queue(loop=loop)
         self._weakref = weakref.ref(self)
@@ -199,8 +201,14 @@ class Connection:
 
     def _create_waiter(self, func_name):
         if self._waiter is not None:
-            raise RuntimeError('%s() called while another coroutine is '
-                               'already waiting for incoming data' % func_name)
+            if self._cancelling:
+                if not self._waiter.done():
+                    raise RuntimeError('%s() called while connection is '
+                                       'being cancelled' % func_name)
+            else:
+                raise RuntimeError('%s() called while another coroutine is '
+                                   'already waiting for incoming '
+                                   'data' % func_name)
         self._waiter = create_future(self._loop)
         return self._waiter
 
@@ -212,6 +220,8 @@ class Connection:
         @asyncio.coroutine
         def cancel():
             self._waiter = create_future(self._loop)
+            self._cancelling = True
+            self._cancellation_waiter = self._waiter
             self._conn.cancel()
             if not self._conn.isexecuting():
                 return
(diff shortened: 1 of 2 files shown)


jettify:
Docker is broken on travis ci, not only in this repo, hopefully they will fix it soon.

tests/test_connection.py
```diff
@@ -288,6 +289,36 @@ def inner():
...
+    fut = asyncio.Future(loop=loop)
+
+    @asyncio.coroutine
+    def inner():
+        fut.set_result(None)
+        yield from cur.execute("SELECT pg_sleep(10)")
+
+    conn = yield from connect()
+    cur = yield from conn.cursor()
+    task = ensure_future(inner(), loop=loop)
```
jettify: Lets pass `cur` and `fut` explicitly as arguments, so it is easier to see where they come from

tests/test_connection.py
```diff
@@ -288,6 +289,36 @@ def inner():
...
+    yield from fut
+    yield from asyncio.sleep(0.1, loop=loop)
+
+    task.cancel()
+
+    for tick in range(100):
+        yield from asyncio.sleep(0, loop=loop)
+        status = conn._conn.get_transaction_status()
```
jettify: `conn` has public method `get_transaction_status`, why you do not use it?

aiopg/connection.py
```diff
@@ -199,8 +201,14 @@ def _fatal_error(self, message):
 
     def _create_waiter(self, func_name):
         if self._waiter is not None:
-            raise RuntimeError('%s() called while another coroutine is '
-                               'already waiting for incoming data' % func_name)
+            if self._cancelling:
+                if not self._waiter.done():
+                    raise RuntimeError('%s() called while connection is '
```
asvetlov: Could you add a test for this line.

themylogin: Done

asvetlov:
Thanks!

PR #265, opened 2017-01-26
add support for "async with cursor_context()"

### LINKED ISSUE #264: cleanup aiopg.pool.Pool.cursor method
right now using this method is not async nice, with the syntax:
```python
        with (yield from pool.cursor()) as cur:
            yield from cur.execute("SELECT 1")
```

Based on the usage of context managers in aiopg, it should instead behavior like aiopg.create_pool, to yield the following usage:
```python
        async with pool.cursor() as cur:
            yield from cur.execute("SELECT 1")
```

In order to accomplish this aiopg.pool.Pool.cursor should not be marked a coroutine, and the acquisition of the connection and cursor should not happen until `__aenter__` of _PoolCursorContextManager.  With the current design probably the easiest way would be to pass a coroutine closure to _PoolCursorContextManager to return the conn and cur during `__aenter__`

thehesiod:
resolves https://github.com/aio-libs/aiopg/issues/264


jettify:
Good point, is it possible to hack something so se can support both: `async with (await pool.cursor())` and `async with pool.cursor()` and then deprecate first example in next release?

thehesiod:
ya I'll work something up in the next few days

thehesiod:
@jettify ok, I think I've done it to have the current method backwards and forwards compatible.  Getting it to work was tricky. Just tweaking the unittests now for code coverage.  Let me know what you think.

thehesiod:
@jettify ping :)

aiopg/utils.py
```diff
@@ -193,37 +197,97 @@ def __aexit__(self, exc_type, exc_val, exc_tb):
 class _PoolCursorContextManager:
```
thehesiod: hmm, wondering if it would have helped inheriting from `_ContextManager`

k4nar:
I'm using this branch in development and it works very well 👍 .

aiopg/utils.py
```diff
@@ -193,37 +197,97 @@ def __aexit__(self, exc_type, exc_val, exc_tb):
...
+
+    @asyncio.coroutine
+    def _yield_await_init(self):
+        yield from self._init_pool(False)
+        self._cur = yield from self._pool.conn.cursor(**self._cursor_kwargs)
+        return self
+
+    def __iter__(self):
+        # This will get hit if you use "yield from pool.cursor()"
+        if PY_35:
```
asvetlov: Don't think we need this warning.
`yield from pool.cursor()` works pretty fine even on Python 3.6.

aiopg/utils.py
```diff
@@ -163,6 +163,10 @@ def __init__(self, pool, conn):
         self._pool = pool
         self._conn = conn
 
+    @property
```
asvetlov: Why do you need this property?

asvetlov (changes requested):

aiopg/utils.py
thehesiod: this is just so they to start using `async with pool.cursor()`  instead of `with (yield from pool.cursor())` so that the cursor will be actually closed at the end of the context.

thehesiod: I guess in the future because right now it doesn't actually wait for the connection to be closed.

aiopg/utils.py
thehesiod: after looking at this more it wouldn't help because we need different behavior between enter and aenter

aiopg/utils.py
thehesiod: used here https://github.com/aio-libs/aiopg/pull/265/files#diff-222246bbd6dde09cb363ee2c57a67887R246

thehesiod:
@jettify I removed the deprecation warning per @asvetlov since it does technically work, we can deprecate it later I suppose.

thehesiod:
ugh, unfortunately don't have enough time to figure out why this fails in travis, gonna revert the docker client upgrade changes :(

thehesiod:
hmm, @asvetlov let me know what you think of this error: https://travis-ci.org/aio-libs/aiopg/jobs/250130251#L1360

thehesiod:
hmm, this was supposed to have been fixed by https://github.com/aio-libs/aiopg/commit/8ca948ec8e8813065483f799187cfc80f04e268d

thehesiod:
ok, looks like you later refactored it in https://github.com/aio-libs/aiopg/commit/2a72b6097e9d0a4c23318b03e57e10767acbbdd4#diff-3416729d6745fac0ca3dd44b22a69068 however it looks like it doesn't work with the latest psycopg2, so I'll let you decide how you want to fix it

tests/pep492/test_async_await.py
```diff
@@ -27,7 +24,26 @@
...
+            await cursor.execute('SELECT 42')
+            resp = await cursor.fetchone()
+            assert resp == (42, )
+        assert cursor.closed
+    assert pool.closed
+
+
+async def test_pool_cursor_await_context_manager(loop, pg_params):
+    async with aiopg.create_pool(loop=loop, **pg_params) as pool:
+        with (await pool.cursor()) as cursor:
```
Daenyth: Should the `yield from` form be tested also?

thehesiod:
@asvetlov ok everything is now passing, let me know if you'd like any other changes

thehesiod:
@asvetlov anything left?  let's get this merged :)

thehesiod:
@vir-mir @asvetlov can we get this merged?

asvetlov:
I can take a look in a few days

thehesiod:
FYI failures not my fault

thehesiod:
@asvetlov @vir-mir ?  :)

PR #303, opened 2017-04-18
Add explicit code blocks for syntax highlighting

maxnordlund:
This is mainly for Github, but should have no effect since the default is python anyway.


jettify:
fixed in other pr

PR #319, opened 2017-05-11
Unnecessary None as second argument

stopspazzing:
By default dict.get() returns None, if the key you were searching for wasn't found. Therefore, you can remove None as second argument.


jettify:
thanks!

PR #361, opened 2017-07-29
Make aiopg installable in PyPy environment

avanov:


asvetlov:
Please enable PyPy in travis config to make sure that library is really functional on PyPy.

avanov:
Is there a list of pypy versions available on Travis CI, or do they follow the same naming scheme as pyenv does? I test the package against `pypy3.5-5.8.0` locally.

asvetlov:
https://docs.travis-ci.com/user/languages/python/#PyPy-Support

avanov:
Ok, currently builds fail on PyPy because of the presence of `psycopg2` in `requirements.txt`. What is the reason for having psycopg2 pinned in `requirements.txt`?

aiopg/utils.py
```diff
@@ -231,3 +233,10 @@ def __exit__(self, *args):
         coroutines._COROUTINE_TYPES += (_ContextManager,)
     except:
         pass
+
+
+# map psycopg2cffi to psycopg2, so any code
+# that imports psycopg2 will use psycopg2cffi
+if PY_IMPL == 'PyPy':
+    from psycopg2cffi import compat
+    compat.register()
```
jettify: I have never used `psycopg2cffi`, but what happened if `compat.register()` called two times? or after first import of `psycopg2` in other module?

avanov: @jettify it import `psycopg2cffi` and makes it available under `psycopg2` name in `sys.modules` (see https://github.com/svpcom/psycopg2cffi/blob/master/psycopg2cffi/compat.py)

As long as`aiopg.utils` is imported in the main package namespace before `psycopg2`, all subsequent occurences of `psycopg2` will be mapped to already imported `psycopg2cffi` (including relative import cases that go through the main `aiopg` namespace anyway).

asvetlov: Hmm. 
Thanks @jettify for pointing on.
Maybe better to explicitly import either `psycopg2` or `psycopg2cffi`?
Doing this we'll never catch an import order problem.

avanov: @asvetlov do you mean you would prefer `psycopg2_compat` module with sth like this inside?

```python
# aiopg/psycopg2_compat

if PY_IMPL == 'PyPy':
    from psycopg2cffi import *
else:
    from psycopg2 import *
```

asvetlov: Yes,

avanov:
@asvetlov shall I remove `psycopg2` from `requirements.txt` ?

jettify:
We use `requirements.txt` to track updated with pyup bot, so can be removed, I believe.

avanov:
The test failures for CPython environments are now consistent with https://github.com/aio-libs/aiopg/pull/362 due to changes introduced to `psycopg2` between version `2.6.2` and `2.7.3`. TravisCI now fetches the most recent psycopg2 version during the `python setup.py install` build step.

steven-upside:
I see that this commit has been approved, are there plans to merge this anytime soon? PyPy + aiopg would be a great combination!

asvetlov:
1. PR was approved by arbitrary github user, not aiopg committer. Basically it is an another form for saying "I like it, +1"
2. Tests are still fails, the Pull Request is not ready for merging yet.

avanov:
@steven-upside so yes, I am waiting for a response from psycopg2cffi on the issue above, and on another one at https://github.com/chtd/psycopg2cffi/issues/85 , after which I think I could finalise this PR.

avanov:
closing this as no longer relevant

PR #366, opened 2017-08-07
Add pool_recycle keyword argument to Pool.

DanCardin:
Add a pool_recycle option similar to the optional available in sqlalchemy

### DIFF
--- a/aiopg/pool.py
+++ b/aiopg/pool.py
@@ -19,12 +19,13 @@ PY_341 = sys.version_info >= (3, 4, 1)
 def create_pool(dsn=None, *, minsize=1, maxsize=10,
                 loop=None, timeout=TIMEOUT,
                 enable_json=True, enable_hstore=True, enable_uuid=True,
-                echo=False, on_connect=None,
+                echo=False, on_connect=None, pool_recycle=-1,
                 **kwargs):
     coro = _create_pool(dsn=dsn, minsize=minsize, maxsize=maxsize, loop=loop,
                         timeout=timeout, enable_json=enable_json,
                         enable_hstore=enable_hstore, enable_uuid=enable_uuid,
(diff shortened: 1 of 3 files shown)


aiopg/pool.py
```diff
@@ -185,6 +193,16 @@ def _fill_free_pool(self, override_min):
         n, free = 0, len(self._free)
         while n < free:
             conn = self._free[-1]
+
+            # Close connections that have lived longer than the recycle limit.
+            if conn in self._free_timestamps:
+                now = datetime.datetime.now()
```
asvetlov: Use `loop.time()` instead

aiopg/pool.py
```diff
@@ -66,9 +68,11 @@ def __init__(self, dsn, minsize, maxsize, loop, timeout, *,
         self._enable_uuid = enable_uuid
         self._echo = echo
         self._on_connect = on_connect
+        self._pool_recycle = pool_recycle
         self._conn_kwargs = kwargs
         self._acquiring = 0
         self._free = collections.deque(maxlen=maxsize or None)
+        self._free_timestamps = {}
```
asvetlov: I suggest storing timestamp into connection object

DanCardin: I was thinking about that, but `Connection` doesn't know whether it's free or not at the moment, that I can tell. You'd be happy with an additional attribute which only gets updated through the pool?

I was going to add a `free` method to Connection to handle updating the time

DanCardin:
Just to note, I've made changes according to the comments/requested changes, though I'm not sure if this is 100% what you meant slash if there are other places or concerns that I've missed which could get the connection out of sync with the pool

DanCardin:
closing because https://github.com/aio-libs/aiopg/pull/373 seems to do the same thing better!

Issue #372, opened 2017-08-28
psycopg2.OperationalError: server closed the connection unexpectedly


soar:
I'm using aiopg with SqlAlchemy support. All your examples looks like:

```python
async with create_engine(...) as engine:
        async with engine.acquire() as conn:
            await conn.execute(...)
```

But I think this is wrong way: you shouldn't `create_engine` on every DB call. So engine object should be reusable instance in whole app. I've tried to use it like this:

```python
# Somewhere on app init stage:
async def init_db_engine():
    db_engine = await aiopg.sa.create_engine(**engine_init_params)
    return db_engine

# Somewhere on app stop stage:
async def release_db_engine(db_engine):
    db_engine.close()
    return await db_engine.wait_closed()

# In some handlers:
async def get(*args, **kwargs):
    redirect_table = app.db.models.redirect.Redirect.__table__
    async with self.application.db.acquire() as conn:
        redirect_query = redirect_table.select().where(...)
        redirect_query_result = await conn.execute(redirect_query)
        redirect_info = await redirect_query_result.fetchone()
```

All looks like good, but... After some time of program inactivity any call to database fails with error:

```
	value = future.result()
	resp = yield from self._coro
	redirect_query_result = await conn.execute(redirect_query)
File "/usr/local/lib/python3.6/site-packages/aiopg/utils.py", line 72, in __await__
File "/usr/local/lib/python3.6/site-packages/aiopg/cursor.py", line 113, in execute
File "/usr/local/lib/python3.6/site-packages/aiopg/connection.py", line 237, in _poll
	yield from self._conn._poll(waiter, timeout)
	yield from cursor.execute(str(compiled), post_processed_params[0])
File "/usr/local/lib/python3.6/site-packages/aiopg/sa/connection.py", line 110, in _execute
	return fut.result()
File "/usr/local/lib/python3.6/asyncio/tasks.py", line 358, in wait_for
File "/usr/local/lib/python3.6/site-packages/aiopg/connection.py", line 134, in _ready
	yield from asyncio.wait_for(self._waiter, timeout, loop=self._loop)

psycopg2.OperationalError: server closed the connection unexpectedly before or while processing the request. This probably means the server terminated abnormally
```

Next call - works fine, so connection is reistablished.

What I've tried:
1. Add parameter `pool_recycle=1` to `engine_init_params` - it is not recognized
2. Add `minsize` and `maxsize` parameters to `engine_init_params` - this didn't help

soar:
[...]

PR #373, opened 2017-08-30
Pool connections recycling

soar:
Related to issue: https://github.com/aio-libs/aiopg/issues/372

Now you can do simply:
```python
db_engine = await aiopg.sa.create_engine(..., recycle=300)
```

... and then all connections which were unused last `300` seconds - will be automatically closed and re-established.

### DIFF
--- a/aiopg/pool.py
+++ b/aiopg/pool.py
@@ -191,7 +191,7 @@ class Pool(asyncio.AbstractServer):
             if conn.closed:
                 self._free.pop()
             elif self._recycle > -1 \
-                    and time.time() - conn.last_usage > self._recycle:
+                    and self._loop.time() - conn.last_usage > self._recycle:
                 conn.close()
                 self._free.pop()
             else:
--- a/aiopg/pool.py
+++ b/aiopg/pool.py
@@ -1,7 +1,6 @@
 import asyncio
 import collections
 import sys
-import time
 import warnings
 
 
--- a/aiopg/pool.py
+++ b/aiopg/pool.py
@@ -1,6 +1,7 @@
 import asyncio
 import collections
 import sys
+import time
 import warnings
 
 
@@ -17,12 +18,12 @@ PY_341 = sys.version_info >= (3, 4, 1)
 
 
 def create_pool(dsn=None, *, minsize=1, maxsize=10,
-                loop=None, timeout=TIMEOUT,
+                loop=None, timeout=TIMEOUT, recycle=-1,
                 enable_json=True, enable_hstore=True, enable_uuid=True,
                 echo=False, on_connect=None,
                 **kwargs):
     coro = _create_pool(dsn=dsn, minsize=minsize, maxsize=maxsize, loop=loop,
-                        timeout=timeout, enable_json=enable_json,
+                        timeout=timeout, recycle=recycle, enable_json=enable_json,
                         enable_hstore=enable_hstore, enable_uuid=enable_uuid,
                         echo=echo, on_connect=on_connect, **kwargs)
     return _PoolContextManager(coro)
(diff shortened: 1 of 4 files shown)


soar:
Also you can see original SQLAlchemy code: https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/pool.py#L623

aiopg/pool.py
```diff
@@ -187,6 +190,10 @@ def _fill_free_pool(self, override_min):
             conn = self._free[-1]
             if conn.closed:
                 self._free.pop()
+            elif self._recycle > -1 \
+                    and time.time() - conn.last_usage > self._recycle:
```
asvetlov: Should be `self._loop.time()`: asyncio uses `time.monotonic()` internally

soar: @asvetlov Done

soar:
@asvetlov I've made changes you requested

DanCardin:
Your PR is very similar to my own, https://github.com/aio-libs/aiopg/pull/366/files, except for *where* you're doing the time setting and resets. Yours has only 1 place where you set and reset the time, and truly if that's sufficient I'm perfectly happy with that.

I'd only suggest you name the option `pool_recycle` to match the option in sqlalchemy's `create_engine` call.

soar:
@DanCardin I thought about `pool_recycle`, but there are too many places to rename this parameter.

DanCardin:
right, but they're all in this PR, right? If you're only talking about renaming things in this PR, I'd be happy to do it. But I feel like maintaining parity between option(s) names definitely has value.

soar:
Ok, @DanCardin - you convinced me.

soar:
@asvetlov We really need this!

jettify:
I will take a look also

aiopg/pool.py
```diff
@@ -17,20 +17,21 @@
 
 
 def create_pool(dsn=None, *, minsize=1, maxsize=10,
-                loop=None, timeout=TIMEOUT,
+                loop=None, timeout=TIMEOUT, pool_recycle=-1,
```
jettify: We need to add check for `pool_recycle` validation, like we do for  `minsize`, also what do you think about making default value `None` or large number, say 3600 seconds?

aiopg/pool.py
```diff
@@ -187,6 +189,10 @@ def _fill_free_pool(self, override_min):
             conn = self._free[-1]
             if conn.closed:
                 self._free.pop()
+            elif self._recycle > -1 \
```
jettify: I think this check a bit fragile, what will happened if someone pass `pool_recycle=-0.5`, I suggest change default and add parameter validation.

jettify:
I like this PR, if my comment addressed I think we can merge. @asvetlov any objections?

aiopg/pool.py
soar: @jettify 
When you have database on localhost - looks like you don't need connection recycling. I wasn't able to reproduce this issue on localhost. So, large values weren't needed by default.

What about `-1` - I use this value only for compatibility with SQLAlchemy. Why did they use `-1` - I don't know.

aiopg/pool.py
soar: I've used code logic from SQLAlchemy for compatibility - see: https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/pool.py#L623

But I also think, that `None` by default and comparision `self._recycle is not None and self._recycle > 0` will be better.

soar:
@jettify @DanCardin 

What do you think about this conversation? I can rewrite code to use `None` by default, but this won't be compatible with SQLAlchemy parameters.

aiopg/pool.py
jettify: make sense, lets leave `-1`

aiopg/pool.py
jettify: agreed

jettify:
@soar lets give @asvetlov few days to look, otherwise I can merge this weekend.

jettify:
thanks!

jettify:
New version available on pypi https://pypi.python.org/pypi/aiopg/0.13.1

soar:
@jettify @asvetlov Thank you!

jettify:
@soar Thanks you for contribution!

PR #398, opened 2017-11-14
Added Logging when connection pool is full and waiting 

him3010:


asvetlov:
Full connection pool is a normal situation, it could happen many times during program lifecicle.
The problem is raised when the pool is full for long time.

Logging every event when the pool becomes full generates too many useless noise.
If you need such information -- use monitoring tools like Prometeus, Graphite/Graphana etc.

PR #412, opened 2017-12-28
Fix SQLAlchemy 1.2.0 compatibility

### LINKED ISSUE #411: Broken compatibility with new release of SQLAlchemy 1.2.0
Hello,

Yesterday released new version of SQLAlchemy (1.2.0) and new release incompatible with aiopg:

```
mymodule.py:42: in fetchone
    result = await conn.execute(query)
.tox/py36-tests/lib/python3.6/site-packages/aiopg/utils.py:72: in __await__
    resp = yield from self._coro
.tox/py36-tests/lib/python3.6/site-packages/aiopg/sa/connection.py:116: in _execute
    return ResultProxy(self, cursor, self._dialect, result_map)
.tox/py36-tests/lib/python3.6/site-packages/aiopg/sa/result.py:234: in __init__
    self._metadata = ResultMetaData(self, cursor.description)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <aiopg.sa.result.ResultMetaData object at 0x7f0417f6a550>
result_proxy = <aiopg.sa.result.ResultProxy object at 0x7f0417f6a5c0>
metadata = (Column(name='column_0', type_code=23, display_size=None, internal_size=4, precision=None, scale=None, null_ok=None), ...Column(name='column_2', type_code=3802, display_size=None, internal_size=-1, precision=None, scale=None, null_ok=None))

    def __init__(self, result_proxy, metadata):
        self._processors = processors = []
    
        result_map = {}
        if result_proxy._result_map:
            result_map = {elem[0]: elem[3] for elem in
                          result_proxy._result_map}
    
        # We do not strictly need to store the processor in the key mapping,
        # though it is faster in the Python version (probably because of the
        # saved attribute lookup self._processors)
        self._keymap = keymap = {}
        self.keys = []
        dialect = result_proxy.dialect
>       typemap = dialect.dbapi_type_map
E       AttributeError: 'PGDialect_psycopg2' object has no attribute 'dbapi_type_map'
```

Gr1N:
Fixes #411

Related SQLAlchemy commit https://github.com/zzzeek/sqlalchemy/commit/31f80b9eaeb3c3435b7f6679b41e434478b1d11c

### DIFF
--- a/aiopg/sa/result.py
+++ b/aiopg/sa/result.py
@@ -100,7 +100,12 @@ class ResultMetaData(object):
         self._keymap = keymap = {}
         self.keys = []
         dialect = result_proxy.dialect
-        typemap = dialect.dbapi_type_map
+
+        # `dbapi_type_map` property removed in SQLAlchemy 1.2+.
+        # Usage of `getattr` only needed for backward compatibility with
+        # older versions of SQLAlchemy.
+        typemap = getattr(dialect, 'dbapi_type_map', {})
+
         assert dialect.case_sensitive, \
             "Doesn't support case insensitive database connection"
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,5 +11,5 @@ pytest-cov==2.5.1
 pytest-sugar==0.9.0
 pytest-timeout==1.2.1
 sphinxcontrib-asyncio==0.2.0
-sqlalchemy==1.1.15
+sqlalchemy==1.2.0
 psycopg2==2.6.2


asvetlov:
Does the patch work with SQLAlchemy 1.1?

Gr1N:
@asvetlov tests are passing with version 1.1, but as you can see in commit https://github.com/zzzeek/sqlalchemy/commit/31f80b9eaeb3c3435b7f6679b41e434478b1d11c SQLAlchemy bumped minimum required version of `cx_Oracle` and removed `dbapi_type_map` for all engines.

aiopg/sa/result.py
```diff
@@ -121,9 +120,7 @@ def __init__(self, result_proxy, metadata):
             name, obj, type_ = (
                 colname,
                 None,
-                result_map.get(
-                    colname,
-                    typemap.get(coltype, sqltypes.NULLTYPE))
```
jettify: We can do `hasattr(dialect, 'dbapi_type_map')` check, and make code working with both version of sqlalchemy

Gr1N: if we want to keep `dbapi_type_map` compatibility, I suggest something like:
```
typemap = getattr(dialect, 'dbapi_type_map', {})
```
and keep `result_map` as is.

Gr1N: @jettify if you prefer this way I can fix my changes

jettify: Looks good, please also add comment explaining why we do that hack

Gr1N: Done, please review

jettify:
LGTM, not sure if we we want to test aiopg with both versions of SQLA, change looks trivial. Anyway I will create ticket for to drop this hack and bump minimal required version for SQLA, after one or two releases, this will give some buffer time for upgrade.

Gr1N:
@jettify could you please upload new version of `aiopg` package to PyPI? I really want to upgrade my projects to SQLAlchemy 1.2 :)

jettify:
already deployed https://pypi.python.org/pypi/aiopg/0.13.2

Gr1N:
Thanks!

jettify:
Just noticed other issue: https://github.com/psycopg/psycopg2/issues/594 we need to bump minimal version of psycopg2, as result half of our builds fail due to this bug.

Issue #364, opened 2017-08-01
close cannot be used while an asynchronous query is underway


ingmferrer:
I'm getting this error if I do multiple fetchs. I'm not sure what I'm doing wrong:
```

Exception ignored in: <function ResultProxy.__init__.<locals>.<lambda> at 0x7f979638fe18>
Traceback (most recent call last):
  File "/home/anthbot/Pyvenv/anthbotv5/lib/python3.5/site-packages/aiopg/sa/result.py", line 235, in <lambda>
    self._weak = weakref.ref(self, lambda wr: cursor.close())
  File "/home/anthbot/Pyvenv/anthbotv5/lib/python3.5/site-packages/aiopg/cursor.py", line 50, in close
    self._impl.close()
psycopg2.ProgrammingError: close cannot be used while an asynchronous query is underway
```
This is the query:
```
query = sa.select([role_item]).where(role_item.c.auto_role_id == auto_role_id)
        return await fetch_ite(query)
```
This is the function:
```
async def _fetch(query, type=DBType.ONE):
    async with engine.acquire() as conn:
        res = await conn.execute(query)
        if type == DBType.ONE:
            return await res.fetchone()
        elif type == DBType.ALL:
            return await res.fetchall()
        elif type == DBType.ITE:
            return res
        return None
```

Any help would be appreciated.

Skorpyon:
Same issue:

DeoLeung:
same here, any hints on what's went wrong?

barrachri:
Same problem here.

the problem is that `self._weak = weakref.ref(self, lambda wr: cursor.close())` try to close the connection when a query is underway (as you read from the traceback.

Strange because that line of code is there from a long time:
https://github.com/aio-libs/aiopg/commit/557ad3c2a0fc27c49cb8f973c7c8ccb47ed4615d#diff-e2e6e38bb0fbb338fc152d8f6b75ca37

barrachri:
This for me resolve the error

```python
if cursor.description is not None:
            self._metadata = ResultMetaData(self, cursor.description)
            self._weak = None
```
Version of the libraries

```python
aiohttp==2.2.5
aiopg==0.13.0
psycopg2==2.7.3.1
SQLAlchemy==1.1.9
SQLAlchemy-Utils==0.32.14
```

Skorpyon:
Any update? Can core developers comment this issue? This exception bore me all time. Please give some advice why its happen and what to do.

[...]

PR #415, opened 2018-01-04
ensure connection is released

thehesiod:
I've seen cases where the close can throw `psycopg2.ProgrammingError: close cannot be used while an asynchronous query is underway` which would previously cause the connection to not be returned to the pool, and cause a connection to "leak" from the pool.  See discussion in: https://github.com/aio-libs/aiopg/issues/364
  
This also fixes running tests on OSX


jettify:
Good point, but should we also close connection in case we have any error on `cursor.close()`, this way pool will recycle it, I am not 100% sure that connection is usable after that exception.

thehesiod:
@jettify but that's what the line that failed tried doing, closing it, and from what I saw on timeouts it tries to cancel the connection as well.

thehesiod:
btw seems to be failing due to something I did not change:
```
    writing manifest file 'pip-egg-info/psycopg2.egg-info/SOURCES.txt'
    Error: could not determine PostgreSQL version from '10.1'
```

jettify:
Regarding CI, I should have fix soon.

jettify:
Failed line tried to do `cursor.close()`? but not `con.close()` or I am missing something?

jettify:
Rebase should fix CI

thehesiod:
btw, let me know if you'd like to handle this via https://docs.python.org/3.5/library/contextlib.html#contextlib.ExitStack, I use it for situations like these to avoid code bloat

jettify:
I like ExitStack too but change is not significant probably not worth it.

jettify:
Regarding change, as you mentioned in other issue, we probably need to close connection, then return is back to the pool. Pool will drop it:
https://github.com/aio-libs/aiopg/blob/b301c9c5c933d7602227fe944d69079b7c7c3ee6/aiopg/pool.py#L250-L261

jettify:
I worry about returning potentially broken connection connection to the pool.

thehesiod:
ok, so then what we need to do is create a new connection for the pool, let me c if I can come up with something

jettify:
no need to create connection, pool should care about that

thehesiod:
with minsize == maxsize == 1:

If we don't call release, and the connection never"closes" we're stuck because there are no "freeable" connections (default recycle is -1).

If we do call release (like this patch does), and the connection never closes, we append a potentially bad connection to the pool.

So I think we need "drop" ability to `Pool.release`

jettify:
I mean close connection and release it back to the pool. Regardless of pool size connection should be always returned, it could be closed but should be returned.

jettify:
drop ability is here:
https://github.com/aio-libs/aiopg/blob/b301c9c5c933d7602227fe944d69079b7c7c3ee6/aiopg/pool.py#L249-L250
we stop tracking connection and do not move it to free connection list

thehesiod:
I'm not sure I'm following what you're saying but I've submitted a change which will allow non-closed connections to be dropped, let me know what you think.

aiopg/pool.py
```diff
@@ -255,10 +258,13 @@ def release(self, conn):
                     tran_status)
                 conn.close()
                 return fut
+
             if self._closing:
                 conn.close()
-            else:
+            elif conn.closed:
```
jettify: Can you elaborate here? bit not follow in what case we need this?

thehesiod: i just realized the fundamental premise of this function is to reuse connections, so there's a problem when we're passing connections that are in a bad state (thus I reverted last change).

Here's the issue which we need to resolve: conn.close() consistently throws an exception, how do we drop it?  Right now we're going to call release, which will add it back to the free pool.  If we don't call release it will be stuck in the used pool.

jettify:
I am thinking more about:
```
    def __exit__(self, *args):
        try:
            self._cur.close()
        except ProperException as e:
            self._conn.close()
        finally:
            try:
                self._pool.release(self._conn)
            finally:
                self._pool = None
                self._conn = None
                self._cur = None

File: aiopg/pool.py (hunk 1 above)
Comment (jettify) [reply aiopg/pool.py]: Hold on but I thought that `self._cur.close()` throws exception in your case not `conn.close()` ?

**Comment by thehesiod** [comment]
ah, I was assuming that if the cursor close fails the connection close would fail as well, but ya I guess not, I'll code that up, thanks

**Comment by jettify** [comment]
Glad to see you find some time to code between your parental responsibilities :)

**Comment by thehesiod** [comment]
thanks, just sent you a link in gitter :)

File: aiopg/utils.py [hunk 2]
```diff
@@ -145,7 +146,7 @@ class _PoolAcquireContextManager(_ContextManager):
     __slots__ = ('_coro', '_conn', '_pool')
 
     def __init__(self, coro, pool):
-        self._coro = coro
+        super().__init__(coro)
```
jettify: not sure about this one, constructor also sets _obj property, it is not listed in slots, probably better have original version

thehesiod: actually, this seems more correct, `_obj` is used for `__aenter__` / `__aexit__` in the parent class, and this is overriding the behavior and thus should be using the same attribute

thehesiod: with even a further simplification :)

jettify:
merged as part of https://github.com/aio-libs/aiopg/pull/421

PR #417, opened 2018-01-05
Change obsolete yield_fixture to just fixture

gyermolenko:


tests/conftest.py
```diff
@@ -169,7 +169,7 @@ def pg_params(pg_server):
     return dict(**pg_server['pg_params'])
 
 
-@pytest.yield_fixture()
+@pytest.fixture()
```
jettify: Can you please review `()` here and in other similar places, better to be consistent everywhere.

jettify:
minore change needed other then that looks good

tests/conftest.py
gyermolenko: done

PR #418, opened 2018-01-05
Upgrade docker-py from 1.10.6 to docker 2.7.0

gyermolenko:
Upgrade docker-py package from 1.10.6 to docker 2.7.0
(original package was renamed)


jettify:
thanks!

PR #421, opened 2018-01-09
Bit better docker 

jettify:
Follow up work for https://github.com/aio-libs/aiopg/pull/415


jettify:
@thehesiod added small fix to your PR here

tests/pep492/test_async_await.py
```diff
@@ -53,6 +54,30 @@
     assert cursor.closed
 
 
+@asyncio.coroutine
+async def test_pool_context_manager_timeout(pg_params, loop):
```
jettify: @thehesiod I reworked test a bit, please review

thehesiod: hmm, wonder how it worked locally, can't believe I checked it in with the debug code still in there, thanks!  looks good!

jettify: Thank! Wanted double check this code.

PR #423, opened 2018-01-19
add connectory cursor factory setter

thehesiod:
per text in http://initd.org/psycopg/docs/connection.html#connection.cursor you're supposed to be able to set the `cursor_factory` on the `Connection` class


asvetlov:
I don't see how http://initd.org/psycopg/docs/connection.html#connection.cursor implies the attribute setter.

The change adds a state to connection object, it might lead to strange behavior if two concurrent tasks change the property.
Why setting the factory on connection creation doesn't satisfy your needs?

thehesiod:
whomever sets it last wins like normal.  This doesn't add any state, it just sets a property that already exists on the underlying object.  Attributes which are read-only are marked as such

jettify:
I agree with @asvetlov such approach could lead to unexpected behavior. For example we have connection pool, and user changes cursor type on one of the connections and forget to set it back to original value. As result other coroutine may acquire connection in one cases with new cursor type in other cases other cursor type in random fashion.

jettify:
I think `conn.cusor(cursor_factory=DictCursor)` is better API since it does not change default cursor.

thehesiod:
IMO it doesn't matter if it's "good" or not, this is an existing psycopg2 API, aiopg is supposed to be an asyncio wrapper around psycopg2, it's not supposed to decide which APIs psycopg2 should or should not expose.  There is already a default, this is just changing the default, it's really not a big deal

asvetlov:
aiopg is not supposed to blindly copy all bad psycopg2 decisions.
aiopg is modeled after psycopg2 but libraries were never strictly 100% compatible.

thehesiod:
this is no different than the `autocommit` setter: https://github.com/aio-libs/aiopg/pull/423/files#diff-4197052d00271d573ed07beab6f61949L416

jettify:
I think we need to drop `autocommit` setter anyway since:
```
There are several limitations in using asynchronous connections: the connection is always in autocommit mode and it is not possible to change it.
```
http://initd.org/psycopg/docs/advanced.html#asynchronous-support

jettify:
Changing default during app work usually leads to unexpected results. Say we have 2 aiohttp handlers one of them uses DictCursor other uses Cursor. If user changes default cursor in one handler at some point other handler will be affected, due to the fact that connection returned to the pool with different default.

asvetlov:
We have enough power to **not add** `cursor_factory` setter, isn't it? :)
Agree with @jettify about dropping `autocommit` setter as well.

thehesiod:
ok lets keep this open then for all the stuff you guys want to drop

Issue #249, opened 2017-01-04
Unable to detect disconnect when using NOTIFY/LISTEN


danielnelson:
I want to ensure that if my database connection goes down that I can recover, but when listening on a notification channel there is no error raised.

I'm using this for testing, it sends a notification every couple seconds.  If you start this and then shutdown the database, the NOTIFY commands reconnect but the listen task is unaware that it has lost its connection.

```
async def notify(pool):
    while True:
        try:
            async with pool.acquire() as conn:
                async with conn.cursor() as cur:
                    await cur.execute("NOTIFY channel")
                    print('NOTIFY channel')
                    await asyncio.sleep(2)
        except Exception as exc:
            print('NOTIFY error %s' % exc)
            await asyncio.sleep(2)


async def listen(pool):
    try:
        async with pool.acquire() as conn:
            async with conn.cursor() as cur:
                await cur.execute("LISTEN channel")
                print('LISTEN channel')
                while True:
                    await conn.notifies.get()
                    print('Notified')
    except Exception as exc:
        print('LISTEN error %s' % exc)


async def main():
    async with aiopg.create_pool(dsn) as pool:
        listener = listen(pool)
        notifier = notify(pool)
        await asyncio.gather(listener, notifier)
    print("ALL DONE")


loop = asyncio.get_event_loop()
loop.run_until_complete(main())
```

RekGRpth:
I solve it so:
```diff
diff --git a/aiopg/connection.py b/aiopg/connection.py
index 8718bec..8086db9 100755
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -160,6 +160,7 @@ class Connection:
                     exc = exc2
             if waiter is not None and not waiter.done():
                 waiter.set_exception(exc)
+            self._notifies.put_nowait(None)
         else:
             if self._fileno is None:
                 # connection closed
```
and code:
```diff
<     try:
<         async with pool.acquire() as conn:
<             async with conn.cursor() as cur:
<                 await cur.execute("LISTEN channel")
<                 print('LISTEN channel')
<                 while True:
<                     await conn.notifies.get()
<                     print('Notified')
<     except Exception as exc:
<         print('LISTEN error %s' % exc)
---
>     while True:
>         try:
>             async with pool.acquire() as conn:
>                 async with conn.cursor() as cur:
>                     await cur.execute("LISTEN channel")
>                     print('LISTEN channel')
>                     while True:
>                         message = await conn.notifies.get()
[...]

PR #431, opened 2018-02-07
Enqueue a marker when no more notifications can possibly be received

gclinch:
This allows a single queue reader to wake up when (for example) the connection is closed.


gclinch:
This should also resolve issue #249

gclinch:
This is better solved by #432

PR #432, opened 2018-02-07
Set an Event when the connection is closed

gclinch:
This allows other waiters (eg waiting on Connection.notifies.get()) to be woken when the connection will never generate more notifications.

Would solve Issue #249 with the use of
`asyncio.wait([connection.closure_event.wait(), connection.notifies.get()], return_when=FIRST_COMPLETED)`

### DIFF
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -116,6 +116,7 @@ class Connection:
         self._writing = False
         self._cancelling = False
         self._cancellation_waiter = None
+        self._closure_event = asyncio.Event(loop=loop)
         self._echo = echo
         self._notifies = asyncio.Queue(loop=loop)
         self._weakref = weakref.ref(self)
@@ -147,6 +148,7 @@ class Connection:
                             # forget a bad file descriptor, don't try to
                             # touch it
                             self._fileno = None
+                        self._closure_event.set()
 
             try:
                 if self._writing:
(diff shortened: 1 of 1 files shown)


gjcarneiro:
This patch didn't solve the problem for me.
I suspect we need something like aioredis' [ClosableQueue](https://github.com/aio-libs/aioredis/blob/master/aioredis/pubsub.py#L397)

gjcarneiro:
Oh, I see, I didn't read carefully enough.  This patch requires API changes.

I think it's not good enough: IMHO, aiopg should raise an exception when waiting for pg notify and the underlying connection got closed.

Requiring the developer to explicitly use the `Event()` object means that most developers will not even be aware of the bug until it bites them in production.

In contrast, an exception raised will most likely be caught somewhere.

gjcarneiro:
I made another PR (#559) that, imho, is a better approach.  It also uses an event, but it is used transparently but the `connection.notifies.get()` coroutine.  Makes life a lot simpler for developers.

vir-mir:
thank. I need a little more time, I plan to start merge into master by the end of next week

gjcarneiro:
No pressure from me, not in a big hurry, I have a workaround for the problem. ;)

Pliner:
Fixed by #559

PR #437, opened 2018-02-14
python3.7 support

smagafurov:


smagafurov:
Problem with python3.7
https://github.com/pypa/setuptools/issues/1257

vir-mir:
@smagafurov you lowered the coverage, could you raise it or leave it as before

smagafurov:
@vir-mir: Sorry but I can't understand how to do this.
See my changes.
May you help me to do this?

smagafurov:
@vir-mir: now code coverage is ok

**!!! but there is some back incompatibility in python 3.7**

This code works in python 3.6
```python
async for value in conn.execute(sql):
```

But in python 3.7 it throws:
```
TypeError: 'async for' received an object from __aiter__ that does not implement __anext__: generator
```

I fix this for python 3.7 as:
```python
async for value in await conn.execute(sql):
```

See my changes in `tests/pep492/test_async_await.py`

May be this solution is wrong and we must do something to support python3.7 without this incompatibility.

tests/pep492/test_async_await.py
```diff
@@ -275,7 +278,10 @@
     result = []
     async with aiopg.sa.create_engine(loop=loop, **pg_params) as engine:
         async with engine.acquire() as conn:
-            async for value in conn.execute(sql):
+            rows = conn.execute(sql)
+            if PY_37:
```
vir-mir: This is a bad decision. We are here solving the consequences and not the cause. you need to rethink it, you probably need to do this in conn.execute

mikhail-yarosh:
any update on this?

asvetlov:
Please postpone the merge.
I need to put my hands on the PR.

aiopg/connection.py
```diff
@@ -106,7 +106,8 @@ class Connection:
 
     def __init__(self, dsn, loop, timeout, waiter, echo, **kwargs):
         self._loop = loop
-        self._conn = psycopg2.connect(dsn, async=True, **kwargs)
```
Se7ge: Here we can use `async_` keyword: http://initd.org/psycopg/docs/connection.html#connection.async_

smagafurov: fixed

Gr1N:
Hi, any updates and estimates on this? Thanks

tests/pep492/test_async_await.py
Ge0: Can anyone else refactor this hack so we can move forward? I don't know what you mean by "doing this in `conn.execute`", @vir-mir.

JustinTArthur: This is a fun one. The examples and documentation show that `aiopg.sa.SAConnection.execute` is both `await`-able and can be asynchronously iterated BUT the current implementation should only work by awaiting result before iterating. It just happens to be working without awaiting the result in CPython < 3.7 because it thinks we were supplying an old-style async iterable and that it should await the iterable first like it did pre-3.5.2.

Now that CPython doesn't revert to awaiting when __aiter__ returns an awaitable, we have a couple choices:
* Change the documentation, examples, tests to emphasize that execute is not async-for-able but its result is, like `async for row in await conn.execute(tbl.select())`

or

* Leave docs, tests, examples as-is, but make our `_SAConnectionContextManager`s behave like they should, with their `__aiter__ `providing something other than a coroutine. This could be the `_SAConnectionContextManager` instance itself, with some `__anext__` implemented, or some other wrapper around the coroutine the manager encapsulates.
  * This might be the best option given folks have probably already been using `for row in conn.execute()` in their code.

JustinTArthur: @smagafurov I've created a possible fix PR to your PR in smagafurov/aiopg#1.

JustinTArthur:
@smagafurov I have two pull requests pending against your PR's branch for fixing some of the remaining issues.

JustinTArthur:
This last one should address the failing CI checks for 3.4.

aiopg/utils.py
```diff
@@ -86,7 +86,20 @@ class _SAConnectionContextManager(_ContextManager):
...
+            def __anext__(self):
+                if self._obj is None:
+                    self._obj = yield from self._coro
+
+                try:
+                    return (yield from self._obj.__anext__())
+                except StopAsyncIteration:
+                    self._obj.close()
+                    self._obj = None
+                    raise
```
JustinTArthur: Ideally, we would just make `__aiter__` an async generator so we don't have to supply this `__anext__` proxying code, but this would break for the small handful of CPython releases >=3.5.2 and <3.6

.travis.yml
```diff
@@ -9,8 +9,10 @@ python:
 - "3.4"
 - "3.5"
 - "3.6"
+- "nightly"
```
smagafurov: Want to replace "nightly" -> "3.7" but there is issue about Travis and python 3.7: https://github.com/travis-ci/travis-ci/issues/9815

webknjaz: You can add a job with specific settings for that

webknjaz: Nevermind. I did this.

webknjaz: Nightly points to 3.8-dev now.

JustinTArthur:
@vir-mir and @asvetlov this should be ready for another pass if you get a moment.

popravich:
LGTM, but I'd rather wait for @asvetlov to share his thoughts on this

vir-mir:
@asvetlov @popravich @jettify if there are no objections, I will wait a few days and salt it.

PR #451, opened 2018-03-13
Added `get_dialect` func to have ability to pass `json_serializer`

Gr1N:
Hi,

I would like to introduce simple change. In my cases I want to use another JSON serializer, for now I need to copy whole dialect initialization code into my sources. I don't want to do this. So in PR you can find new function called `get_dialect`. Changes backward compatible, but I'm open to make any changes and improvements.

Best regards


Gr1N:
@vir-mir could you please release new version of package?

vir-mir:
@Gr1N no, I do not release releases

@asvetlov  when the new release?

PR #498, opened 2018-08-22
sqlalchemy adapter trx begin allow transaction_mode

shanecarey17:
Allows transaction mode to be set on `SAConnection.begin()` method calls


vir-mir:
LGTM - I'll salt in a couple of days.

PR #534, opened 2019-01-28
Project edits for pr517

vir-mir:
- Modification in tests
- Removing old style file examples
- Removing old style (yield from) in docs
- Rename psycopg2 to psycopg2-binary in docs
- added note in docs about going to python3.5.2+
- Interim fix due to issues #533 
- refactor docs style
- aded the example code 

@asvetlov @jettify since the changes are not compatible with previous versions, I propose to upgrade to major version `1.0.0`

PR #548, opened 2019-03-23
resolve issues 535 and 364

vir-mir:
Dear all.

I think we hurried using one cursor to connect.
I learned how the connection works. `Cursors are not thread safe: a multithread application can create many cursors from the same connection and should use each cursor from a single thread. See Thread and process safety for details.`  http://initd.org/psycopg/docs/usage.html#thread-safety

This quote may be misleading:
`Connections shouldn’t be shared either by different green threads: see Support for coroutine libraries for further details.` this only applies to `greenlet` more info here (https://github.com/psycopg/psycopg2/blob/8b7506f80d62366ab6bb9e055ff4ea492d16db2c/psycopg/pqpath.c#L830)

**aiopg** use async inteface (https://github.com/psycopg/psycopg2/blob/8b7506f80d62366ab6bb9e055ff4ea492d16db2c/psycopg/pqpath.c#L874) 

issues #364 in that we do not close the cursor, at the time of release for the connection here (https://github.com/aio-libs/aiopg/blob/master/aiopg/sa/engine.py#L244) and here (https://github.com/aio-libs/aiopg/blob/master/aiopg/utils.py#L132)

Solving the problem, was the closure of cursors at the time of release of the connection.
I also track weak references in the `SAConnection` class to release the cursor that closes before it is released connection.

this is necessary in order to clear cursors for users who do not reconfigure the dialect (https://github.com/aio-libs/aiopg/blob/master/aiopg/sa/engine.py#L39)

The default settings are such that `insert` and `update` requests will return to the result. what does not allow to close the cursor.

surprisingly all the tests were successful. and also this example works successfully https://github.com/aio-libs/aiopg/issues/364#issuecomment-373445693

### DIFF
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -120,7 +120,6 @@ class Connection:
         self._cancelling = False
         self._cancellation_waiter = None
         self._echo = echo
-        self._cursor_instance = None
         self._notifies = asyncio.Queue(loop=loop)
         self._weakref = weakref.ref(self)
         self._loop.add_reader(self._fileno, self._ready, self._weakref)
@@ -284,13 +283,6 @@ class Connection:
     async def _cursor(self, name=None, cursor_factory=None,
                       scrollable=None, withhold=False, timeout=None):
 
-        if not self.closed_cursor:
-            warnings.warn(('You can only have one cursor per connection. '
-                           'The cursor for connection will be closed forcibly'
-                           ' {!r}.').format(self), ResourceWarning)
-
-        self.free_cursor()
-
         if timeout is None:
             timeout = self._timeout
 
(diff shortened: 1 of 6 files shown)


vir-mir:
I also added a dialect test. This will be described in the documentation here https://github.com/aio-libs/aiopg/pull/534.

```python
async def test_dialect_returning(make_engine, connect):
    dialect = get_dialect()
    dialect.implicit_returning = False

    engine = await make_engine(dialect=dialect)
        res = await conn.execute(tbl.insert().values(id='1', name='test'))
        assert res.closed

        res = await conn.execute(
            tbl.insert().values(id='2', name='test').returning(tbl.c.id)
        )
        assert not res.closed
        assert await res.scalar() == '2'
        assert res.closed
```

vir-mir:
I will be grateful if you have @runtel will time to check.

thehesiod:
IIRC the prime issue was that you can't use two cursors with active SQL commands on same asynchronous connection as you'll get a complaint from the underlying PostgreSQL library

thehesiod:
Per http://initd.org/psycopg/docs/advanced.html: Two cursors can’t execute concurrent queries on the same asynchronous connection.

vir-mir:
This statement is true only for parallel cursors in one connection. async in python this competitiveness, which means opening a new cursor in one connection is correct. the main thing is not to overeat it in parallel processes. I will try to add a test in order to show it clearly.

Pliner:
Resolved by #801

PR #552, opened 2019-04-01
Fix on_connect multiple call on acquire

### LINKED ISSUE #551: Pool on_connect callback called on every acquire
When using an `on_connect` callback, it appears to get called every time a connection is aquired, rather than every time a new connection is made, which appears to differ from the stated documentation:

> on_connect – a callback coroutine executed at once for every created connection. May be used for setting up connection level state like client encoding etc.

Although it looks like it was intended to put one of 'least' or 'most' between the 'at' and 'once'?

This is due to the callback check being made in the acquire code: https://github.com/aio-libs/aiopg/blob/9fdf7b92b4a705fcffe259af70b80a0afaf7bed7/aiopg/pool.py#L165-L180

Should this instead be located after the new connection is made in `_fill_free_pool`: https://github.com/aio-libs/aiopg/blob/master/aiopg/pool.py#L199

The test case [here](https://github.com/aio-libs/aiopg/blob/9fdf7b92b4a705fcffe259af70b80a0afaf7bed7/tests/test_pool.py#L551) currently doesn't catch this, as it only checks the call is made once. Perhaps it should instead check that when reusing a connection, the `on_connect` callback isn't called multiple times, maybe by using a pool with `minsize = maxsize = 1` and acquiring the connection twice?

aaliddell:
Fixes #551 

The Pool `on_connect` handler is now called exactly once per connection, rather than on every call to `acquire`

Also, some things on your PR template checklist can't be done, as you don't have a CONTRIBUTORS.txt or a CHANGES folder?

## Checklist

- ~~If you provide code modification, please add yourself to `CONTRIBUTORS.txt`~~
- ~~Add a new news fragment into the `CHANGES` folder~~

### DIFF
--- a/aiopg/pool.py
+++ b/aiopg/pool.py
@@ -173,8 +173,6 @@ class Pool(asyncio.AbstractServer):
                     assert not conn.closed, conn
                     assert conn not in self._used, (conn, self._used)
                     self._used.add(conn)
-                    if self._on_connect is not None:
-                        await self._on_connect(conn)
                     return conn
                 else:
                     await self._cond.wait()
@@ -203,6 +201,8 @@ class Pool(asyncio.AbstractServer):
                     enable_uuid=self._enable_uuid,
                     echo=self._echo,
                     **self._conn_kwargs)
+                if self._on_connect is not None:
+                    await self._on_connect(conn)
                 # raise exception if pool is closing
                 self._free.append(conn)
                 self._cond.notify()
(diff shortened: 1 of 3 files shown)


brianmaissy:
+1 on this, can we get it merged?

alxpy:
Can you please merge this patch? PR is small and has test :+1: 

cc @asvetlov @popravich @vir-mir (based on MAINTAINERS.txt)

aaliddell:
Is this fix actually going to be considered for merge if I go through the CLA process?

Pliner (approved):

Pliner:
Hi @aaliddell,

Thanks for the contribution.

PR #559, opened 2019-04-09
Fix "Unable to detect disconnect when using NOTIFY/LISTEN", Closes #249

### LINKED ISSUE #249: Unable to detect disconnect when using NOTIFY/LISTEN

gjcarneiro:
## What do these changes do?

When connection to DB is closed, any tasks pending on `await connection.notifies.get()` receives an exception, instead of hanging forever.

Now `connection.notifies.get()` can raise a psycopg2 exception, whereas before this would hang forever.

## Related issue number

#249

## Checklist

- [x] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is `<Name> <Surname>`.
  * Please keep alphabetical order, the file is sorted by names. 

### DIFF
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -30,7 +30,12 @@ import psycopg2.extensions
 import psycopg2.extras
 
 from .log import logger
-from .utils import _ContextManager, create_completed_future, get_running_loop
+from .utils import (
+    ClosableQueue,
+    _ContextManager,
+    create_completed_future,
+    get_running_loop,
+)
 
 TIMEOUT = 60.0
 
@@ -762,6 +767,7 @@ class Connection:
         self._writing = False
         self._echo = echo
         self._notifies = asyncio.Queue()  # type: ignore
+        self._notifies_proxy = ClosableQueue(self._notifies)
         self._weakref = weakref.ref(self)
         self._loop.add_reader(
             self._fileno, self._ready, self._weakref  # type: ignore
@@ -806,6 +812,7 @@ class Connection:
                     # chain exception otherwise
                     exc2.__cause__ = exc
                     exc = exc2
+            self.notifies.close(exc)
             if waiter is not None and not waiter.done():
                 waiter.set_exception(exc)
         else:
@@ -1182,9 +1189,9 @@ class Connection:
             self._loop.call_exception_handler(context)
 
     @property
-    def notifies(self) -> asyncio.Queue:  # type: ignore
-        """Return notification queue."""
-        return self._notifies
+    def notifies(self) -> ClosableQueue:
+        """Return notification queue (an asyncio.Queue -like object)."""
+        return self._notifies_proxy
 
     async def _get_oids(self) -> Tuple[Any, Any]:
         cursor = await self.cursor()
--- a/aiopg/utils.py
+++ b/aiopg/utils.py
@@ -122,3 +122,53 @@ class _IterableContextManager(_ContextManager[_TObj]):
             finally:
                 self._obj = None
             raise
+
+
+class ClosableQueue:
+    """
+    Proxy object for an asyncio.Queue that is "closable"
+
+    When the ClosableQueue is closed, with an exception object as parameter,
+    subsequent or ongoing attempts to read from the queue will result in that
+    exception being result in that exception being raised.
+
+    Note: closing a queue with exception will still allow to read any items
+    pending in the queue.  The close exception is raised only once all items
+    are consumed.
+    """
+
+    def __init__(self, queue: asyncio.Queue):
+        self._queue = queue
+        self._close_exception = asyncio.Future()  # type: asyncio.Future[None]
+
+    def close(self, exception: Exception) -> None:
+        if not self._close_exception.done():
+            self._close_exception.set_exception(exception)
+
+    async def get(self) -> Any:
+        loop = get_running_loop()
+        get = loop.create_task(self._queue.get())
+
+        _, pending = await asyncio.wait(
+            [get, self._close_exception],
+            return_when=asyncio.FIRST_COMPLETED,
+        )
+
+        if get.done():
+            return get.result()
+        get.cancel()
+        self._close_exception.result()
+
+    def empty(self) -> bool:
+        return self._queue.empty()
+
+    def qsize(self) -> int:
+        return self._queue.qsize()
+
+    def get_nowait(self) -> Any:
+        try:
+            return self._queue.get_nowait()
+        except asyncio.QueueEmpty:
+            if self._close_exception.done():
+                self._close_exception.result()
+            raise
--- a/examples/notify.py
+++ b/examples/notify.py
@@ -1,5 +1,7 @@
 import asyncio
 
+import psycopg2
+
 import aiopg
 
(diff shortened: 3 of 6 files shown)


gjcarneiro:
What is this checklist item "Add a new news fragment into the CHANGES folder"?  There is no CHANGES folder, as far as I can see...

vir-mir:
ignore, I will correct later

michaelbeaumont:
@vir-mir hi! any news on this PR?

Pliner:
@vir-mir I'm interested in the fix too. Could you review it, please? :wink:

albertodonato:
any news on this PR? I would be interested in the fix too.

@gjcarneiro it seems there are a few conflicts, could you please update the branch?

examples/notify.py
```diff
@@ -18,7 +19,11 @@ async def listen(conn):
     async with conn.cursor() as cur:
         await cur.execute("LISTEN channel")
         while True:
-            msg = await conn.notifies.get()
+            try:
+                msg = await conn.notifies.get()
+            except psycopg2.Error as ex:
+                print("ERROR: ", ex)
+                return
```
albertodonato, line 26: the `try/except` here could be replaced with a `with contextlib.suppress(psycopg2.Error)` (and the `print` dropped)

gjcarneiro:
I've updated the PR for current master.  Can we get this merged please?  Been 2 years... :/

Pliner:
Hi @gjcarneiro,

Sorry for the so long delay. I hope I will be able to review it today or tomorrow. Thanks for rebasing it 👍

Pliner:
@gjcarneiro I'm a bit concerned about one aspect of the implementation: it covers only get method of the queue(I suspect not so many people use get_no_wait(for instance), but anyway). 

What do you think about exposing a kind of proxy object which exposes only async get method(probably we need to expose any other methods/properties)?

gjcarneiro:
> @gjcarneiro I'm a bit concerned about one aspect of the implementation: it covers only get method of the queue(I suspect not so many people use get_no_wait(for instance), but anyway).

That's a good point. I didn't think about `get_nowait()`, it would probably also need to check for the psycopg exception, for consistency.

> What do you think about exposing a kind of proxy object which exposes only async get method(probably we need to expose any other methods/properties)?

Well, I tried to make this 100% backwards compatible.  What you propose, I don't know if it would be 100% compatible?  It would probably be cleaner, but...  I'll think about it.

gjcarneiro:
I've updated the PR with making ClosableQueue a proxy class instead of a subclass.

aiopg/utils.py
```diff
@@ -122,3 +122,46 @@ async def __anext__(self) -> _TObj:
...
+    def empty(self) -> bool:
+        return self._queue.empty()
+
+    def qsize(self) -> int:
+        return self._queue.qsize()
+
+    def get_nowait(self) -> Any:
+        if self._close_exception.done():
+            self._close_exception.result()
+        return self._queue.get_nowait()
```
Pliner, line 167: Behaviour of `get_nowait` differs from `get`: `get` allows to get all received notifications in case of the closed connection, but `get_nowait` throws exception right after a closure of the connection.

aiopg/connection.py
```diff
@@ -806,6 +812,7 @@ def _ready(weak_self: "weakref.ref[Any]") -> None:
                     # chain exception otherwise
                     exc2.__cause__ = exc
                     exc = exc2
+            self.notifies.close(exc)
```
Pliner, line 815: What do you think? Is it important here to include an exception info from psycopg to the ClosableQueue? Instead we could raise QueueIsClosedError or smth like this.

aiopg/utils.py
Pliner: BTW I don't know which behaviour is better 🤣 From the first glance a behaviour of `get` is more convenient.

aiopg/connection.py
Pliner: It's just a question, I'm ok with current implementation.

aiopg/utils.py
gjcarneiro: I'm not sure you are correct about `get()`. I hadn't thought about it.  What happens when there is an error but there are pending items in the queue?  :thinking: 

As far as I can tell, it's kind of undefined whether get() returns the next item or we raise the exception.  Because we are asking asyncio to wait for two futures, and they will both be ready at the same time, here:

```python
        _, pending = await asyncio.wait(
            [get, self._close_exception],
            return_when=asyncio.FIRST_COMPLETED,
        )
```
Does the `get` future become completed immediately, or does it need to be given a chance to execute by the main loop before becoming completed?  It's probably semi-random, which one is returned, to be honest.

If you think it matters, I can work some more on this to make sure that while there are items pending in the queue we keep consuming them before raising the exception.

aiopg/connection.py
gjcarneiro: > What do you think? Is it important here to include an exception from psycopg to the ClosableQueue? Instead we could raise QueueIsClosedError or smth like this.

My feeling is that there might be some information in the psycopg2 exception that may be useful to the application to determine _why_ the connection got broken?  I think it's slightly more useful to keep the original exception.  I am leaning towards keeping as is.

gjcarneiro: I am choosing to pass the original exception as is.

aiopg/utils.py
gjcarneiro: I've updated the code and docs so that both `get()` and `get_nowait()` keep consuming items until the queue is empty, and only check for the close exception when the queue becomes empty.

gjcarneiro:
I've made a new version.  Any more comments?

gjcarneiro:
ping

Pliner:
Sorry for the delay, quite overloaded at work, hope will be able review it on this week.

albertodonato:
Awesome! Could you please make a release with this fix?

Pliner:
> Awesome! Could you please make a release with this fix?

Hi @albertodonato,

Surething, but it will be the beta version(probably `1.3.0b4`) to double-check that everything works. I'm going to deploy this release later today.

Pliner:
https://pypi.org/project/aiopg/1.3.0b4/

Issue #275, opened 2017-02-10
Using "localhost" as host rather than "127.0.0.1 blocks connections


mikn:
Using the example code from the documentation:
```import asyncio
import aiopg

dsn = 'dbname=aiopg user=aiopg password=passwd host=localhost'

async def go():
    async with aiopg.create_pool(dsn) as pool:
        async with pool.acquire() as conn:
            async with conn.cursor() as cur:
                await cur.execute("SELECT 1")
                ret = []
                async for row in cur:
                    ret.append(row)
                assert ret == [(1,)]

loop = asyncio.get_event_loop()
loop.run_until_complete(go())
```
Results in this:
```
$ time python pgtest.py 
Traceback (most recent call last):
  File "pgtest.py", line 17, in <module>
    loop.run_until_complete(go())
  File "/usr/lib/python3.5/asyncio/base_events.py", line 466, in run_until_complete
    return future.result()
  File "/usr/lib/python3.5/asyncio/futures.py", line 293, in result
    raise self._exception
  File "/usr/lib/python3.5/asyncio/tasks.py", line 241, in _step
    result = coro.throw(exc)
  File "pgtest.py", line 7, in go
    async with aiopg.create_pool(dsn) as pool:
  File "/home/mikaelk/venvs/csc-backend/lib/python3.5/site-packages/aiopg/utils.py", line 77, in __aenter__
    self._obj = yield from self._coro
  File "/home/mikaelk/venvs/csc-backend/lib/python3.5/site-packages/aiopg/pool.py", line 46, in _create_pool
    yield from pool._fill_free_pool(False)
  File "/home/mikaelk/venvs/csc-backend/lib/python3.5/site-packages/aiopg/pool.py", line 203, in _fill_free_pool
    **self._conn_kwargs)
  File "/home/mikaelk/venvs/csc-backend/lib/python3.5/site-packages/aiopg/utils.py", line 67, in __iter__
    resp = yield from self._coro
  File "/home/mikaelk/venvs/csc-backend/lib/python3.5/site-packages/aiopg/connection.py", line 74, in _connect
    yield from conn._poll(waiter, timeout)
  File "/home/mikaelk/venvs/csc-backend/lib/python3.5/site-packages/aiopg/connection.py", line 239, in _poll
    yield from asyncio.shield(cancel(), loop=self._loop)
  File "/usr/lib/python3.5/asyncio/futures.py", line 380, in __iter__
    yield self  # This tells Task to wait for completion.
[...]

PR #577, opened 2019-06-27
Fixing connection with multiple hosts and unavailable replicas

r-dmv:
## What do these changes do?

Since PostgreSQL 10 it's possible to describe multiple hosts in connection string, it is useful when you have HA cluster of PostgreSQL with several replicas and you don't want to have some kind of balancer over your replicas.   

And know it works in `aiopg` if you're using libpq >= 10. 

But if there is a dead or unavailable replica in your connection string before your target host, exception is raised after timeout: ```psycopg2.OperationalError: asynchronous connection attempt underway```.
It is also happens when you have host with multiple ip addresses in DNS, and first one is not responding. 

You can simply reproduce this behavior  by setting up 1 PostgreSQL server on 127.0.0.1:5432 and try to connect with connection string:
`dbname=<db> user=<user> password=<password> host=8.8.8.8,127.0.0.1 port=5433,5432 target_session_attrs=read-write connect_timeout=1`

In sync psycopg2 connection it  works properly, but this code:

```
import asyncio
import aiopg

dsn = 'dbname=<db> user=<user> password=<password> host=8.8.8.8,127.0.0.1 port=5433,5432 target_session_attrs=read-write connect_timeout=11'


async def go():
    async with aiopg.create_pool(dsn, timeout=3) as pool:
                await cur.execute("SELECT 1")
                ret = []
                async for row in cur:
                    ret.append(row)
                assert ret == [(1,)]

loop.run_until_complete(go())
```

raises ```psycopg2.OperationalError: asynchronous connection attempt underway```.

This happens because libpq creates new connection, and closes failed, but we are waiting a failed one in `_poll` method and fails after timeout. 

No

## Related issue number

#275 

## Checklist

- [] Add a new news fragment into the `CHANGES` folder

### DIFF
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -28,6 +28,9 @@ TIMEOUT = 60.0
 # to OSError.errno EBADF
 WSAENOTSOCK = 10038
 
+# Connection status from psycopg2 (psycopg2/psycopg/connection.h)
+CONN_STATUS_CONNECTING = 20
+
 
 async def _enable_hstore(conn):
     cur = await conn.cursor()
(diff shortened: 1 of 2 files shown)


r-dmv:
This fix is wrong! 
It fixes not all cases. I'll redo this.

PR #579, opened 2019-07-11
Fixing connection with multiple hosts and unavailable replicas

r-dmv:
## What do these changes do?

Since PostgreSQL 10 it's possible to describe multiple hosts in connection string, it is useful when you have HA cluster of PostgreSQL with several replicas and you don't want to have some kind of balancer over your replicas.   

And know it works in `aiopg` if you're using libpq =< 10. 

But if there is a dead or unavailable replica in your connection string before your target host, exception is raised after timeout: ```psycopg2.OperationalError: asynchronous connection attempt underway```.
It is also happens when you have host with multiple ip addresses in DNS, and first one is not responding. 

You can simply reproduce this behavior  by setting up 1 PostgreSQL server on 127.0.0.1:5432 and try to connect with connection string:
`dbname=<db> user=<user> password=<password> host=8.8.8.8,127.0.0.1 port=5433,5432 target_session_attrs=read-write connect_timeout=1`

In sync psycopg2 connection it  works properly, but this code:

```
import asyncio
import aiopg

dsn = 'dbname=<db> user=<user> password=<password> host=8.8.8.8,127.0.0.1 port=5433,5432 target_session_attrs=read-write connect_timeout=1'


async def go():
    async with aiopg.create_pool(dsn, timeout=3) as pool:
                await cur.execute("SELECT 1")
                ret = []
                async for row in cur:
                    ret.append(row)
                assert ret == [(1,)]

loop.run_until_complete(go())
```

raises ```psycopg2.OperationalError: asynchronous connection attempt underway```.

This happens because libpq creates new connection, and closes failed, but we are waiting a failed one in `_poll` method and fails after timeout. 

Now `aiopg` uses  `connect_timeout` param from `dsn`. 
It is used to timeout connection to the single host, in order to prevent getting stuck on connection to host which is not responding at all, even on first `SYN` packet. 

So now, if you have 3 hosts in your connection string, you should provide `connect_timeout` = `timeout / 3`  just as guarantee that `aiopg` will try all 3 host during the timeout. 

```
# we will wait single host for 1 second and whole connect operation will take 3 seconds at worst case
pool = aiopg.create_pool('host=localhost ... connect_timeout=1', timeout=3)
```
You can also use `connect_timeout` in `kwargs`:

```
pool = aiopg.create_pool('host=localhost ...', timeout=3, connect_timeout=1)
```

## Related issue number

#275 

## Checklist

- [] Add a new news fragment into the `CHANGES` folder

### DIFF
--- a/aiopg/connection.py
+++ b/aiopg/connection.py
@@ -3,10 +3,12 @@ import contextlib
 import errno
 import platform
 import select
+import socket
 import sys
 import traceback
 import warnings
 import weakref
+import ctypes
 
 import psycopg2
 from psycopg2 import extras
@@ -27,6 +29,17 @@ TIMEOUT = 60.0
 # Windows specific error code, not in errno for some reason, and doesnt map
 # to OSError.errno EBADF
 WSAENOTSOCK = 10038
+
+# Connection status from psycopg2 (psycopg2/psycopg/connection.h)
+CONN_STATUS_CONNECTING = 20
+
+# In socket.socket we should know type and family to shutdown socket by fd
+# This function is used for shutdown libpq connection
+# where family and type is unknown
+libc = ctypes.CDLL(None)
+socket_shutdown = libc.shutdown
+socket_shutdown.restypes = ctypes.c_int
+socket_shutdown.argtypes = ctypes.c_int, ctypes.c_int
 
 
 async def _enable_hstore(conn):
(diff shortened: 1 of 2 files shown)


aiopg/connection.py
```diff
@@ -124,10 +142,26 @@ def __init__(self, dsn, loop, timeout, waiter, echo, **kwargs):
...
             self._source_traceback = traceback.extract_stack(sys._getframe(1))
 
+    @staticmethod
+    def _shutdown(weak_self):
+        # Make sure that we won't get stuck in a blocking read or write
+        # inside poll, then give libpq a chance to try another host.
+        # If there is an error, we'll get it from poll.
+        self = weak_self()
+
+        if self._conn.status == CONN_STATUS_CONNECTING:
```
hhell: This occasionally fails due to `self` being `None`.

Minimal fix would be to skip the rest when that is the case,

but generally there shouldn't be active timers for a deleted object.

aiopg/connection.py
```diff
@@ -28,6 +30,17 @@
 # to OSError.errno EBADF
 WSAENOTSOCK = 10038
 
+# Connection status from psycopg2 (psycopg2/psycopg/connection.h)
+CONN_STATUS_CONNECTING = 20
+
+# In socket.socket we should know type and family to shutdown socket by fd
+# This function is used for shutdown libpq connection
+# where family and type is unknown
```
webknjaz, line 38: What's the problem with just using the socket object?

PR #632, opened 2019-12-05
added extras execute_batch for executemany

vir-mir:
I added this function, since it allows you to quickly insert values ​​into the database.

This solves the `executemany` problem, more details can be found in this  issues https://github.com/psycopg/psycopg2/issues/491

I wrote several of my performance tests that showed the following result.
insert 10000  query in database

test first https://github.com/vir-mir/kdr2019/tree/master/test2
aiopg not execute_batch:  18.12 sec
asyncpg: 10.09 sec

test second https://github.com/vir-mir/kdr2019/tree/master/test3
aiopg execute_batch:  0.64 sec
asyncpg: 10.09 sec

### DIFF
--- a/aiopg/extras.py
+++ b/aiopg/extras.py
@@ -1,0 +1,43 @@
+def _paginate(seq, page_size):
+    """Consume an iterable and return it in chunks.
+
+    Every chunk is at most `page_size`. Never return an empty chunk.
+    """
+    page = []
+    count = len(seq)
+    it = iter(seq)
+    for s in range(count + 1):
+        try:
+            for i in range(page_size):
+                page.append(next(it))
+            yield page
+            page = []
+        except StopIteration:
+            if page:
+                yield page
+            return
+
+
+async def execute_batch(cur, sql, argslist, page_size=100):
+    r"""Execute groups of statements in fewer server roundtrips.
+
+    Execute *sql* several times, against all parameters set (sequences or
+    mappings) found in *argslist*.
+
+    The function is semantically similar to
+
+    .. parsed-literal::
+
+        *cur*\.\ `~cursor.executemany`\ (\ *sql*\ , *argslist*\ )
+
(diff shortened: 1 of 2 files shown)


iAnanich:
When this could be merged? Can't wait to use it )

aiopg/extras.py
```diff
@@ -0,0 +1,43 @@
+def _paginate(seq, page_size):
```
berenda-andrey:
```
import itertools 


def _paginate(seq, page_size):
    it = iter(seq)
    page = list(itertools.islice(it, 0, page_size))
    while page:
        yield page
        page = list(itertools.islice(it, 0, page_size))
```
I think it is better solution

berenda-andrey: with this solution you can you iterators in _paginate
for example 
`_paginate(range(1, 8), 3)`

tests/test_extras.py
```diff
@@ -0,0 +1,54 @@
...
+def cursor(connect, loop):
+    async def go():
+        return await (await connect()).cursor()
+
+    cur = loop.run_until_complete(go())
+    yield cur
+    cur.close()
+
+
+def test__paginate():
```
berenda-andrey:
```
@pytest.mark.parametrize(
    'data, expected_pages',
    (
            (
                    (1, 2, 3, 4, 5, 6, 7), [
                        [1, 2, 3],
                        [4, 5, 6],
                        [7],
                    ]
            ),
            (
                    (1, 2, 3, 4, 5, 6), [
                        [1, 2, 3],
                        [4, 5, 6],
                    ]
            ),
    )
)
def test__paginate(data, expected_pages):
    for page, expected_page in itertools.zip_longest(
            _paginate(data, page_size=3),
            expected_pages,
    ):
        assert page == expected_page
```

vir-mir: thanks!

PR #487, opened 2018-07-11
Fix closing of transaction

decaz:
In case of errors within a transaction (for example, connection errors) connection can't be released due to a transaction is still considered active. So we should guarantee deactivating of transaction in case of commit or rollback.

Fixes #332.

PR #666, opened 2020-03-20
Invalidate incorrectly released connections in Engine. closes #665

### LINKED ISSUE #665: An exception in BEGIN does not free up a slot in pool
This is likely the same problem as https://github.com/aio-libs/aiopg/pull/487

When an exception on `BEGIN` happens, the connection is left in a broken state, taking up a slot in the pool indefinitely.

Reproduction script (that monkeypatches forced `raise` on `BEGIN`): https://gist.github.com/hhell/0b9aa6759065bd0310f29f3d0ae4c63d

The script hangs on 15th `acquire`, but should complete successfully (considering the `except` catch).

hhell:
Fix for #665 

See also: #487

### DIFF
--- a/aiopg/__init__.py
+++ b/aiopg/__init__.py
@@ -3,9 +3,10 @@ import sys
 import warnings
 from collections import namedtuple
 
-from .connection import connect, Connection, TIMEOUT as DEFAULT_TIMEOUT
+from .connection import TIMEOUT as DEFAULT_TIMEOUT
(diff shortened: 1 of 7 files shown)


Pliner:
Hi @hhell,

Thanks for the contribution. 

The same issue was resolved by #756 and released as part of https://github.com/aio-libs/aiopg/releases/tag/v1.1.0.

PR #722, opened 2020-08-21
Update required python version

ffe4:
## What do these changes do?

Change the warning about supported python versions to be consistent with the required version on PyPI

see: https://github.com/aio-libs/aiopg/blob/master/setup.py#L61

no

## Related issue number

none

## Checklist

- [ ] I think the code is well written
- [ ] Unit tests for the changes exist


Pliner:
Hi @ffe4,

Thanks for the contribution. 

It looks like I've fixed merely the same thing in #753.

PR #753, opened 2020-12-02
Travis to GA

Pliner:
The aim of the PR is to migrate from Travis to GA as it is now.

There are two separate things to do:
1. Bump psycopg2 to latest and fix tests
2. Add python 3.8 and 3.9 to CI and fix tests

PS Also it seems to be ok to drop python 3.5 support.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python }}
+    - name: Install dependencies
+      run: |
+        pip install -e .
+        pip install -Ur requirements.txt
```
asvetlov, line 35: Not sure if we should use `-U` here, it can lead to unpredictable errors.
Usually, I follow the schema:
1. `requirements.txt` has strict versions only (`dep-name==1.2.0`).
2. `pip install -r requirements.txt` installs only specified versions
3. `dependabot` creates PR for each dependency update, you can verify if the CI passes the version bump correctly and merge the PR.

asvetlov (commented):
Looks good, a minor change is desirable.

.github/workflows/ci.yml
Pliner: Thanks for the comment, I've fixed it.

asvetlov (approved):
Cool, please merge

PR #754, opened 2020-12-04
Bump psycorg and sqlalchemy, add python 3.8, 3.9 to test matrix

Pliner:
Min supported version of psycorg is 2.8.4 because of python 3.8

### DIFF
--- a/tests/test_async_await.py
+++ b/tests/test_async_await.py
@@ -1,6 +1,5 @@
 import asyncio
 
-import psycopg2
 import pytest
 
 import aiopg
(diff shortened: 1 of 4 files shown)


tests/test_async_await.py
```diff
@@ -56,7 +55,7 @@ async def test_pool_context_manager_timeout(pg_params, loop):
     async with aiopg.create_pool(**pg_params, minsize=1,
                                  maxsize=1) as pool:
         cursor_ctx = await pool.cursor()
-        with pytest.raises(psycopg2.ProgrammingError):
+        with pytest.warns(ResourceWarning):
```
Pliner, line 58: @asvetlov I'm pretty unsure about this, but it looks like this behaviour has changed because of psycorg update to 2.8+. Do you have any thoughts about this?

asvetlov: I don't monitor psycopg2 changelist but this line looks good to me if the test passes.

Pliner: Cool, thanks.

PS I had tried to figure out why it could happened from change log, but failed to do so 😢

PR #558, opened 2019-04-08
Fixes Engine.release method to release connection in any way

vmagamedov:
Fixes #332 error.

The problem is that `Engine.release` method raises `InvalidRequestError` exception before releasing connection. This may happen not only when user did something wrong, but also when connection becomes broken. Pool becomes exhausted and application stops responding.

I think that it doesn't matter who is guilty, connection should be released in any way, and `Pool.release` also checks connection's state, and if it is not in IDLE state, pool closes connection.

I've added `test_release_broken_connection` to reproduce this error. Currently `Pool.release` emits `ResourceWarning`, I think that it is enough.

PR #756, opened 2020-12-05
Fix Engine.release method to release connection in any way

### LINKED ISSUE #332: Error when canceling during transaction begin.
I'm seeing an issue when my coroutine is cancelled while trying to begin a transaction:
```python
async with self.db.replica.acquire() as conn, conn.begin():
    res = await conn.execute(query)
    return await res.scalar()
```
```python
  File "/home/project/.pex/install/project-0+unknown-py3-none-any.whl.fd4fcf00c6491ee26d0a46ece96c5eca887e87a0/project-0+unknown-py3-none-any.whl/project/core/app.py", line 235, in get_current_version
    async with self.db.replica.acquire() as conn, conn.begin():
  File "/home/project/.pex/install/aiopg-0.13.0-py3-none-any.whl.2df26f3f2b931863ced66bc4d8059fd21bba4006/aiopg-0.13.0-py3-none-any.whl/aiopg/utils.py", line 77, in __aenter__
    self._obj = yield from self._coro
  File "/home/project/.pex/install/aiopg-0.13.0-py3-none-any.whl.2df26f3f2b931863ced66bc4d8059fd21bba4006/aiopg-0.13.0-py3-none-any.whl/aiopg/sa/connection.py", line 171, in _begin
    yield from self._begin_impl()
  File "/home/project/.pex/install/aiopg-0.13.0-py3-none-any.whl.2df26f3f2b931863ced66bc4d8059fd21bba4006/aiopg-0.13.0-py3-none-any.whl/aiopg/sa/connection.py", line 180, in _begin_impl
    yield from cur.execute('BEGIN')
  File "/home/project/.pex/install/aiopg-0.13.0-py3-none-any.whl.2df26f3f2b931863ced66bc4d8059fd21bba4006/aiopg-0.13.0-py3-none-any.whl/aiopg/cursor.py", line 113, in execute
    yield from self._conn._poll(waiter, timeout)
  File "/home/project/.pex/install/aiopg-0.13.0-py3-none-any.whl.2df26f3f2b931863ced66bc4d8059fd21bba4006/aiopg-0.13.0-py3-none-any.whl/aiopg/connection.py", line 240, in _poll
    raise exc
  File "/home/project/.pex/install/aiopg-0.13.0-py3-none-any.whl.2df26f3f2b931863ced66bc4d8059fd21bba4006/aiopg-0.13.0-py3-none-any.whl/aiopg/connection.py", line 237, in _poll
    yield from asyncio.wait_for(self._waiter, timeout, loop=self._loop)
  File "/opt/python3.6/lib/python3.6/asyncio/tasks.py", line 345, in wait_for
    yield from waiter
concurrent.futures._base.CancelledError

During handling of the above exception, another exception occurred:

  File "/home/project/.pex/install/project-0+unknown-py3-none-any.whl.fd4fcf00c6491ee26d0a46ece96c5eca887e87a0/project-0+unknown-py3-none-any.whl/project/core/app.py", line 354, in middleware
    return await handler(request)
  File "/home/project/.pex/install/project-0+unknown-py3-none-any.whl.fd4fcf00c6491ee26d0a46ece96c5eca887e87a0/project-0+unknown-py3-none-any.whl/project/core/app.py", line 385, in middleware
    response = await handler(request)
  File "/home/project/.pex/install/project-0+unknown-py3-none-any.whl.fd4fcf00c6491ee26d0a46ece96c5eca887e87a0/project-0+unknown-py3-none-any.whl/project/authentication/__init__.py", line 42, in middleware
    return await handler(request)
  File "/home/project/.pex/install/project-0+unknown-py3-none-any.whl.fd4fcf00c6491ee26d0a46ece96c5eca887e87a0/project-0+unknown-py3-none-any.whl/project/health.py", line 37, in health_check
    'data_version': await request.app.get_current_version(),
  File "/home/project/.pex/install/project-0+unknown-py3-none-any.whl.fd4fcf00c6491ee26d0a46ece96c5eca887e87a0/project-0+unknown-py3-none-any.whl/project/core/app.py", line 237, in get_current_version
    return await res.scalar()
  File "/home/project/.pex/install/aiopg-0.13.0-py3-none-any.whl.2df26f3f2b931863ced66bc4d8059fd21bba4006/aiopg-0.13.0-py3-none-any.whl/aiopg/utils.py", line 140, in __aexit__
    yield from self._pool.release(self._conn)
  File "/home/project/.pex/install/aiopg-0.13.0-py3-none-any.whl.2df26f3f2b931863ced66bc4d8059fd21bba4006/aiopg-0.13.0-py3-none-any.whl/aiopg/sa/engine.py", line 169, in release
    raise InvalidRequestError("Cannot release a connection with "
aiopg.sa.exc.InvalidRequestError: Cannot release a connection with not finished transaction 
```

It seems like the issue occurs when the [cursor coroutine](https://github.com/aio-libs/aiopg/blob/ba3a3424680215be28033b03c180a0683028dbf9/aiopg/sa/connection.py#L178) is interrupted. This is after [the `_transaction` is set on the SA connection](https://github.com/aio-libs/aiopg/blob/ba3a3424680215be28033b03c180a0683028dbf9/aiopg/sa/connection.py#L170), but since the error is in the `_TransactionContextManager`'s `__aenter__`, its `__aexit__` is never called to issue a rollback (oddly the `__aexit__` seems to be called successfully if the error is in the `try: finally:` block below this. I don't understand how this happens). Since we have a transaction, and it's not "finished" (although technically it was never started), the engine doesn't realize it can release the connection.

Pliner:
The goal of the PR is to actualise the fix made in #558 by @vmagamedov, but with slightly reworked tests.

Fixes #332, #665, #666 as well.

It looks like both exceptions(`psycopg2.InterfaceError` and `psycopg2.OperationalError`) could be raised, see https://github.com/aio-libs/aiopg/pull/756/checks?check_run_id=1503163697, initially only the first one was expected in the tests.

Tests are written using intermediate tcp proxy to allow disrupt a connection and cover two cases:
1. When a connection has disrupted before a request(`SELECT 1;` for instance) is executed, so rollback could not be executed. 
2. When a connection has disrupted before begin is executed.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -391,3 +391,76 @@ def warning():
 @pytest.fixture
 def log():
     yield _AssertLogsContext
+
+
+@pytest.fixture
+def tcp_proxy(loop):
+    proxy = None
+
+    async def go(src_port, dst_port):
+        nonlocal proxy
+        proxy = TcpProxy(
+            dst_port=dst_port,
+            src_port=src_port,
+        )
+        await proxy.start()
+        return proxy
+    yield go
+    if proxy is not None:
+        loop.run_until_complete(proxy.disconnect())
+
+
+class TcpProxy:
+    """
+    TCP proxy. Allows simulating connection breaks in tests.
+    """
+    MAX_BYTES = 1024
+
+    def __init__(self, *, src_port, dst_port):
+        self.src_host = '127.0.0.1'
+        self.src_port = src_port
+        self.dst_host = '127.0.0.1'
+        self.dst_port = dst_port
+        self.connections = set()
+
+    async def start(self):
+        return await asyncio.start_server(
+            self.handle_client,
+            host=self.src_host,
+            port=self.src_port,
+        )
+
+    async def disconnect(self):
+        while self.connections:
+            writer = self.connections.pop()
+            writer.close()
+            await writer.wait_closed()
+
+    @staticmethod
+    async def _pipe(
+        reader: asyncio.StreamReader, writer: asyncio.StreamWriter
+    ):
+        try:
+            while not reader.at_eof():
+                bytes_read = await reader.read(TcpProxy.MAX_BYTES)
+                writer.write(bytes_read)
+        finally:
+            writer.close()
+
+    async def handle_client(
+        self,
+        client_reader: asyncio.StreamReader,
+        client_writer: asyncio.StreamWriter,
+    ):
+        server_reader, server_writer = await asyncio.open_connection(
+            host=self.dst_host,
+            port=self.dst_port
+        )
+
+        self.connections.add(server_writer)
+        self.connections.add(client_writer)
+
+        await asyncio.wait([
+            self._pipe(server_reader, client_writer),
+            self._pipe(client_reader, server_writer),
+        ])
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -434,7 +434,8 @@ class TcpProxy:
         while self.connections:
             writer = self.connections.pop()
             writer.close()
-            await writer.wait_closed()
+            if hasattr(writer, "wait_closed"):
+                await writer.wait_closed()
 
     @staticmethod
     async def _pipe(
--- a/tests/test_sa_engine.py
+++ b/tests/test_sa_engine.py
@@ -189,4 +189,5 @@ async def test_release_after_connection_disconnected_before_begin(
                 await tcp_proxy.disconnect()
                 async with conn.begin():
                     pytest.fail("Should not be here")
+
     assert engine.size == 0
--- a/tests/test_sa_engine.py
+++ b/tests/test_sa_engine.py
@@ -1,5 +1,6 @@
 import asyncio
 
+import psycopg2
 import pytest
 from psycopg2.extensions import parse_dsn
 from sqlalchemy import Column, Integer, MetaData, String, Table
@@ -80,10 +81,10 @@ def test_not_context_manager(engine):
 async def test_release_transacted(engine):
     conn = await engine.acquire()
     tr = await conn.begin()
-    with pytest.raises(sa.InvalidRequestError):
-        engine.release(conn)
+    with pytest.warns(ResourceWarning, match='Invalid transaction status'):
+        await engine.release(conn)
     del tr
-    await conn.close()
+    assert conn.closed
 
 
 def test_timeout(engine):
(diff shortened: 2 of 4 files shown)


tests/test_sa_engine.py
```diff
@@ -147,3 +148,46 @@ async def test_terminate_with_acquired_connections(make_engine):
...
+    server_port = pg_params["port"]
+    proxy_port = unused_port()
+
+    tcp_proxy = await tcp_proxy(proxy_port, server_port)
+    engine = await make_engine(port=proxy_port)
+
+    with pytest.raises(
+        (psycopg2.InterfaceError, psycopg2.OperationalError)
+    ):
+        with pytest.warns(ResourceWarning, match='Invalid transaction status'):
```
vmagamedov, line 165: Is it ok to spam users with both warning and exception? Maybe it is better just to raise an exception? Connection reset is not a user's fault, so I think that this warning doesn't help them.

tests/test_sa_engine.py
```diff
@@ -80,10 +81,10 @@ def test_not_context_manager(engine):
 async def test_release_transacted(engine):
     conn = await engine.acquire()
     tr = await conn.begin()
-    with pytest.raises(sa.InvalidRequestError):
-        engine.release(conn)
+    with pytest.warns(ResourceWarning, match='Invalid transaction status'):
+        await engine.release(conn)
```
vmagamedov, line 85: Maybe it is better to raise an exception here instead of a warning. Looks like a user's bug to release a connection in the middle of transaction.

tests/test_sa_engine.py
Pliner: Thanks for noticing this. 

I'm a newbie in the aiopg codebase and don't know the reasons why these warnings are reported in such a way.

@asvetlov Could you share your opinion please?

tests/conftest.py
```diff
@@ -391,3 +391,77 @@ def warning():
...
+            self.handle_client,
+            host=self.src_host,
+            port=self.src_port,
+        )
+
+    async def disconnect(self):
+        while self.connections:
+            writer = self.connections.pop()
+            writer.close()
+            if hasattr(writer, "wait_closed"):
```
Pliner, line 437: In python 3.6 there is no such a method

tests/test_sa_engine.py
asvetlov: My initial thought was preventing a programming error and signal this situation early.
I agree that the connection releasing can be done for network reasons, there no need for extra checks.
I see two options:
1. Drop exception in 'release' as PR does, https://github.com/aio-libs/aiopg/blob/master/aiopg/pool.py#L242-L248 warning also can be dropped. It can lead to reports about unfinished transactions by **next** acquired connection.
Satisfactory but confusing.
2. Raise an exception only if the underlying socket is not closed but swallow the transaction status for disconnected peers. It prevents false positives from network errors but can report a programming error fast. Connection for the unclosed transaction should be dropped immediately, the connection is in an undefined incorrect state and cannot be recovered. Perhaps a warning generated in the pool can be replaced with a strict error also.

The 1) is easier, 2) is harder to implement but maybe more correct.
I don't want to decide what to do; I have no time to contribute anyway.  Please choose what you want.

Pliner: Okay, got you, thanks.

I've opened a new [issue](https://github.com/aio-libs/aiopg/issues/761) to fix it a bit later, because it looks like merging(testing) of the PR with incorrect warnings in quite rare case is much better than suffering from broken pool.

tests/test_sa_engine.py
Pliner: I've added this comment to the [issue](https://github.com/aio-libs/aiopg/issues/761) as well.

tests/test_sa_engine.py
asvetlov: Makes sense

PR #770, opened 2020-12-20
Fix IsolationLevel.read_comitted and introduce IsolationLevel.default

### LINKED ISSUE #699: The default read committed isolation level is the wrong choice
Hello, 

I have a project where I experienced serialization failures on a statement:

```
select 1 from file where (id, group_id) = (%s, %s) for update
```

resulting in errors like:

```
psycopg2.errors.SerializationFailure: could not serialize access due to concurrent update
```

now, I happen to know that the statement above would have orderly blocked in case of concurrency, if run in psycopg or in psql: logged some of the operations of the database and sure enough found:

```
postgres_1    | 2020-07-16 10:30:36.377 UTC [11914] LOG:  statement: BEGIN ISOLATION LEVEL REPEATABLE READ
```

Which led me to find:

https://github.com/aio-libs/aiopg/blob/3fb3256319de766384b4867c7cc6710397bd1a8c//aiopg/cursor.py#L10-L16

this is pretty wrong: the driver is choosing an isolation level that causes side effects, and the behaviour of the program becomes very different from the one it would have in psycopg2.

The right default would be to begin a transaction just with `BEGIN`. Note that, unlike what stated here:

https://github.com/aio-libs/aiopg/blob/3fb3256319de766384b4867c7cc6710397bd1a8c/aiopg/transaction.py#L51-L55

`BEGIN` is not "read committed" behaviour: this is "use the server configured behaviour", which is what selected by `default_transaction_isolation` level, and can be changed at server configuration level, session level.

So, what I think is:

1) The `ReadCommittedCompiler` should actually be implemented as:

```python
class ReadCommittedCompiler(IsolationCompiler):
    name = 'Read committed'

    def begin(self):
        return 'BEGIN ISOLATION LEVEL READ COMMITTED'
```

2) you need a class

```python
class DefaultCompiler(IsolationCompiler):
    name = 'Default'

    def begin(self):
        return 'BEGIN'
```

3) that class should be the default.

```python
 class Cursor: 
     def __init__(self, conn, impl, timeout, echo): 
         # ...
         self._transaction = Transaction(self, IsolationLevel.default)
```

you also probably want to fix #497 to allow the isolation level to be chosen, but at least by default you should leave the server default unaltered.

Pliner:
PR introduces `IsolationLevel.default` and fixes `IsolationLevel.read_comitted` as it was mentioned by @dvarrazzo.

Instead of hacking
```
        with (await pool.cursor()) as cur:
            # Set the connection in read committed isolation level
            # to avoid serialization failures on file lock
            # https://github.com/aio-libs/aiopg/issues/699
            # You can unlock the aiopg version from setup.py when this hack
            # is no more needed.
            cur._transaction = aiopg.Transaction(
                cur, aiopg.IsolationLevel.read_committed
            )
            yield cur
```
PR allows to pass `IsolationLevel`
```
        with (await pool.cursor(isolation_level=aiopg.IsolationLevel.read_committed)) as cur:
            yield cur
```

## Related issue number

Fixes #699

## Checklist

- [X] Unit tests for the changes exist

### DIFF
--- a/aiopg/transaction.py
+++ b/aiopg/transaction.py
@@ -1,7 +1,7 @@
 import enum
 import uuid
 import warnings
-from abc import ABC, abstractmethod
+from abc import ABC
 
 import psycopg2
 
@@ -11,12 +11,16 @@ __all__ = ('IsolationLevel', 'Transaction')
 
 
 class IsolationCompiler(ABC):
-    __slots__ = ()
+    __slots__ = ('_isolation_level', '_readonly', '_deferrable')
+
+    def __init__(self, isolation_level, readonly, deferrable):
+        self._isolation_level = isolation_level
+        self._readonly = readonly
+        self._deferrable = deferrable
 
     @property
-    @abstractmethod
     def name(self):
-        ...  # pragma: no cover
+        return self._isolation_level
 
     def savepoint(self, unique_id):
         return 'SAVEPOINT {}'.format(unique_id)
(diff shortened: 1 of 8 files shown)


aiopg/transaction.py
```diff
@@ -77,10 +94,26 @@ def begin(self):
         return query
 
 
+class DefaultCompiler(IsolationCompiler):
+    __slots__ = ()
+
+    def __init__(self, readonly, deferrable):
+        if readonly or deferrable:
+            raise ValueError("Readonly or deferrable are not supported")
```
dvarrazzo, line 102: Why readonly and deferrable are not supported? As this object is running `BEGIN` it should be able to support it, they are independent from the isolation level. 

https://www.postgresql.org/docs/current/sql-begin.html

dvarrazzo: I think it was right that `__slots__` was on the base class and that these properties are available to all the hierarchy

Pliner: image

I've only changed the current implementation 🙈

Pliner: >Why readonly and deferrable are not supported? As this object is running BEGIN it should be able to support it, they are independent from the isolation level.

I dunno why it was unsupported, but let me check

Pliner: image

https://www.postgresql.org/docs/13/sql-set-transaction.html

It looks like readonly supported for any isolation level, but deferred - only for serialisable with readonly.

But I'm a but unsure about current behaviour with raise of an exception.

dvarrazzo: You shouldn't impose a rule on top of what postgres implements.

If it's a no-op it may be a no-op only for the current implementation and change in the future. If the database accepts it, it is not the driver's business to decide to raise an exception.

This seems to work: the database doesn't raise exceptions or even warnings:

```
piro=# begin isolation level read uncommitted deferrable;
BEGIN
*piro=#
```

so I don't see a reason to special-case it and raise an exception.

Pliner: Agree, thanks 🎉

PR #570, opened 2019-06-03
Remove blocking connection.cancel() method

brianmaissy:
Closes #491 
Closes #375

PR #778, opened 2020-12-23
Don't run ROLLBACK when the connection is closed.

### LINKED ISSUE #777: Don't rollback a transaction if the connection is already closed
Caused by #570, in which we close the connection in any situation where we would have cancelled the query, since we don't have an async way of cancelling the query.

One possible situation in which we could find ourselves closing the connection is in the case where a query timed out or was cancelled (in an asyncio sense). In that case, we now close the connection, but the calling code may be inside a transaction context manager, which will try to emit a `ROLLBACK` before exiting the block, since an exception was raised.

The net effect is that instead of getting a `TimeoutError` as excepted, the user gets a `psycopg2.InterfaceError: connection already closed` exception, which gets raised during the handling of the `TimeoutError`. In this case, since we were the ones who closed the connection, and we are also the ones trying to run the `ROLLBACK`, it doesn't seem fair to the user not to deal with it properly (as opposed to if a user runs some cleanup query after a `TimeoutError` was thrown, in which case it would be the user's responsibility to check that the connection is still open).

There are many possible ways to solve this, but I think the most elegant might be just not to run the `ROLLBACK` if the connection is closed. This is nice because it isn't just a workaround for a bug, semantically it makes sense. If the connection is closed, the transaction is over, and there's no need to run a `ROLLBACK`.

brianmaissy:
Fixes #777.

This can be caused when a query times out while running, for example, and the connection is closed as a result (as opposed to cancelling the query, since PR #570). In this case, we would rather not emit the ROLLBACK (the connection is already closed, so the transaction is over anyway), rather than raising an exception when trying to use a connection which is already closed.

I wasn't sure if it was cleaner to do this here (inside the implementation of rollback) or in the context manager of the transaction. The disadvantage of this implementation is that here we are also affecting any rollback that someone does manually, but implementing it in the context manager is awkward since it doesn't have access to the connection object, and we wouldn't have a way to set `self._transaction = None` and `self._is_begin = False`.

If you think the side effect of making all ROLLBACKs on a closed connection silently do nothing is fine, we can leave it like this. Otherwise maybe we can add another intermediate method `rollback_unless_closed` or something like that which the context manager calls, which does the check and calls rollback if necessary.

### DIFF
--- a/aiopg/sa/connection.py
+++ b/aiopg/sa/connection.py
@@ -258,7 +258,7 @@ class SAConnection:
 
     async def _rollback_to_savepoint_impl(self, name, parent):
         if self._connection.closed:
-            self._transaction = None
+            self._transaction = parent
             return
 
         cur = await self._get_cursor()
--- a/aiopg/sa/connection.py
+++ b/aiopg/sa/connection.py
@@ -211,6 +211,10 @@ class SAConnection:
             self._transaction = None
 
     async def _rollback_impl(self):
+        if self._connection.closed:
+            self._transaction = None
+            return
+
         cur = await self._get_cursor()
         try:
             await cur.execute('ROLLBACK')
@@ -253,6 +257,10 @@ class SAConnection:
             cur.close()
 
     async def _rollback_to_savepoint_impl(self, name, parent):
+        if self._connection.closed:
+            self._transaction = None
+            return
+
         cur = await self._get_cursor()
         try:
             await cur.execute(f'ROLLBACK TO SAVEPOINT {name}')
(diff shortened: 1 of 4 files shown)


Pliner (commented):
Could you please add the same test for asyncio.CancelledError?

brianmaissy:
@Pliner have you had a chance to go over this again?

Pliner:
> @Pliner have you had a chance to go over this again?

Sure, I have. I'm a bit unsure about the way how it's resolved in PR, let me try to propose alternative one in a couple of days.

brianmaissy:
@Pliner bump?

Pliner (approved):

Pliner:
@brianmaissy thanks for your contribution and sorry for the long delay. I've slightly modified PR and it is merged to master now.

I'm going to create new dev version today.

Pliner:
https://pypi.org/project/aiopg/1.2.0b3/

PR #801, opened 2021-03-21
Multiple cursors support

### LINKED ISSUE #535: Inner async for loop caused cursor already closed error
After upgrade to `aiopg==0.16.0` next pseudo-code caused `cursor already closed` `InterfaceError`,

```py
async def view(request: web.Request) -> web.Response:
    async with request.app['db'].acquire() as conn:
         data = await fetch_data(conn)
    return web.json_response(convert_data(data))

async def fetch_data(conn: SAConnection) -> List[RowProxy]:
     result: List[RowProxy] = []
     async for item in conn.execute(query):
          result.extend(await fetch_inner_data(conn, item))
     return result

async def fetch_data_item(conn: SAConnection, item: RowProxy) -> List[RowProxy]:
     result: List[RowProxy] = []
     async for item in conn.execute(query):
          result.append(query)
     return result
```

The exception itself is:

```
InterfaceError: cursor already closed
  ...
  File "/api/views.py", line 388, in view
    context.user)
  File "/api/storage.py", line 403, in fetch_data
    async for item in conn.execute(query):
  File "aiopg/utils.py", line 97, in __anext__
    return (yield from self._obj.__anext__())
  File "aiopg/sa/result.py", line 359, in __anext__
    ret = yield from self.fetchone()
  File "aiopg/sa/result.py", line 401, in fetchone
    row = yield from self._cursor.fetchone()
  File "asyncio/coroutines.py", line 120, in coro
    res = func(*args, **kw)
  File "aiopg/cursor.py", line 194, in fetchone
    ret = self._impl.fetchone()
```

The error happened when running `aiohttp.web` server, as well as in tests, run by `pytest`. However downgrade to `aiopg==0.15.0` fixes the issue. Any ideas on why it is happened and maybe you need more debug information?

OS: macOS 10.14.3, Ubuntu 16.04.3
Python version: 3.7.2
aiopg version: 0.16.0

Pliner:
## What do these changes do?

Support of multiple cursors which was initially proposed in #548 to resolve #535 and #364 by @vir-mir. I've slightly modified the implementation and rebased on top of master.

## Related issue number

#777, #535, #364, #548, #778

## Checklist

- [X] Unit tests for the changes exist


aiopg/pool.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, dsn, minsize, maxsize, timeout, *,
             raise ValueError("maxsize should be not less than minsize")
         self._dsn = dsn
         self._minsize = minsize
-        self._loop = get_running_loop(kwargs.pop('loop', None) is not None)
```
Pliner, line 46: I suppose we are ready to drop this

aiopg/sa/engine.py
```diff
@@ -224,18 +227,15 @@ class _ConnectionContextManager:
...
     def __enter__(self):
         return self._conn
 
     def __exit__(self, *args):
-        try:
-            self._engine.release(self._conn)
-        finally:
-            self._engine = None
-            self._conn = None
+        asyncio.ensure_future(self._conn.close(), loop=self._loop)
```
Pliner, line 240: close is coroutine, so we have to await it somehow

aiopg/utils.py
```diff
@@ -88,8 +70,13 @@ async def __aenter__(self):
         return self._obj
 
     async def __aexit__(self, exc_type, exc, tb):
-        self._obj.close()
-        self._obj = None
+        try:
+            if asyncio.iscoroutinefunction(self._obj.close):
+                await self._obj.close()
```
Pliner, line 75: `obj.close` could be a coroutine, for instance, in case of `Engine.acquire`.

PR #806, opened 2021-03-23
Fix compatibility with SA 1.4 for IN statement

### LINKED ISSUE #798: aiopg 1.1.0+ is incompatible with SQLAlchemy 1.4
Aiopg will automatically download SQLAlchemy 1.4.0(Released March 15) which leads to crazy errors due to incompatibility. Make sure to freeze SQLAlchemy==1.3.23.

Two specific issues i noticed:
1. Failure when doing a select. What i found that the query is absolutely valid and Postgres returns what is expected, but the RowProxy fails to map the columns.
```
self = <[InvalidRequestError("Ambiguous column name 'None' in result set! try 'use_labels' option on select statement.") raised in repr()] RowProxy object at 0x7fccab231460>
key = 'brands_id'

     def __getitem__(self, key):
         try:
>           processor, obj, index = self._keymap[key]
E           KeyError: 'brands_id'

/usr/local/lib/python3.7/site-packages/aiopg/sa/result.py:27: KeyError

During handling of the above exception, another exception occurred:

tables = None, sa_engine = <aiopg.sa.engine.Engine object at 0x7fccab2813d0>
```

2. Failure when Deleting from table. Honestly i'm not sure if aiopg is responsible for that, it may very well be psycopg2.
```
E           psycopg2.errors.SyntaxError: syntax error at or near "["
E           LINE 1: ...ETE FROM materials WHERE materials.product_id IN ([POSTCOMPI...
E                                                                        ^

/usr/local/lib/python3.7/site-packages/aiopg/connection.py:106: SyntaxError
```

AVOstap:
Fixes #798

Now in SA, there are two options for how to compile the query.
`literal_binds` and `render_postcompile`
Both options work correctly with the current code base since `aiopg` does not support executemany
https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#all-in-expressions-render-parameters-for-each-value-in-the-list-on-the-fly-e-g-expanding-parameters

Unfortunately, this solution does not support late-expanded
https://docs.sqlalchemy.org/en/14/changelog/migration_12.html#late-expanded-in-parameter-sets-allow-in-expressions-with-cached-statements
to support this feature need to rework the `_execute` method of the `SAConnection` class

### DIFF
--- a/aiopg/sa/connection.py
+++ b/aiopg/sa/connection.py
@@ -103,7 +103,8 @@ class SAConnection:
         if isinstance(query, str):
             await cursor.execute(query, dp)
         elif isinstance(query, ClauseElement):
-            compiled = query.compile(dialect=self._dialect, compile_kwargs=self._query_compile_kwargs)
+            compiled = query.compile(dialect=self._dialect,
+                                     compile_kwargs=self._query_compile_kwargs)
             # parameters = compiled.params
             if not isinstance(query, DDLElement):
                 if dp and isinstance(dp, (list, tuple)):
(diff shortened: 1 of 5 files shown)


Pliner (approved):

Pliner:
Thanks 👍

Issue #807, opened 2021-03-23
aiopg1.2 incompatible with peewee-async


inyutin:
Removing ```kwargs.pop('loop', None)``` in [Multiple cursors support](https://github.com/aio-libs/aiopg/pull/801/files#) make peewee-async, library that used aiopg, incompatible. Probably, some other libraries affected too.

The thing is: before 1.2 you was allow to pass loop in kwargs.  Before passing those kwargs to psycopg2, the loop parameter was popped. Now, it is not. Loop is passing to psycopg2 and we get error:
```
psycopg2.ProgrammingError: invalid dsn: invalid connection option "loop"
```

Pliner:
https://pypi.org/project/aiopg/1.2.1/

PR #808, opened 2021-03-23
pop loop in connection init due to backward compatibility

inyutin:
## What do these changes do?

This is fix to https://github.com/aio-libs/aiopg/issues/807


Pliner:
@inyutin Thanks for your contribution and sorry for this backward incompatible change

Issue #314, opened 2017-05-09
Add type annotations


Daenyth:
It would help in using this library in a codebase with mypy, and since this is a python3-only project, you can use the inline syntax rather than type comments.

jettify:
Sure, eventually I think most projects will add, mypy annotations. Would you like to contribute stub files?

Daenyth:
I can probably do some here and there, though a full codebase-wide tagging effort would be difficult. I'll see what I can do

Pliner:
Resolved by #813 and available as [1.3.0b2](https://pypi.org/project/aiopg/1.3.0b2/)

PR #813, opened 2021-04-02
Type annotations

Pliner:
## What do these changes do?

This PR adds type hints for aiopg except `aiopg.sa`. Due to cyclic dependencies between cursor, connection and transaction, I had to merge three files into one. The same thing is with context managers: due to cyclic dependencies(context managers were very specific) I had to generalise as much as possible `_ContextManager` and move specific ones closer to their usage.

## Related issue number

#314

## Checklist

- [ ] Unit tests for the changes exist


aiopg/connection.py
```diff
@@ -35,22 +56,646 @@ def connect(dsn=None, *, timeout=TIMEOUT, enable_json=True,
...
+        return Connection(
+            dsn,
+            timeout,
+            bool(echo),
+            enable_hstore=enable_hstore,
+            enable_uuid=enable_uuid,
+            enable_json=enable_json,
+            **kwargs
+        )
+    return _ContextManager[Connection](connect_(), disconnect)
```
Pliner, line 69: `_ContextManager` expects a coroutine

spaceorc (approved):

PR #859, opened 2021-07-05
Fix ClosableQueue.get on cancellation, close it on Connection.close

Pliner:
We had the following crutch to force reconnection, it was necessary before the fix #559.

```
async with connection.cursor() as cursor:
    await cursor.execute(f"LISTEN {self._channel}")
    while True:
        try:
            async with async_timeout.timeout(self._reconnect_timeout):
                message = await connection.notifies.get()
                yield message.payload
        except asyncio.TimeoutError:
            logger.info("Too long delay between notifications. We have to reconnect")
            break
```

It helped to find the issue in `ClosableQueue` implementation: `ClosableQueue.get` doesn't handle cancellation well, `get loop.create_task(self._queue.get())` is not cancelled if a cancellation happens in `await asyncio.wait([get, self._close_exception], return_when=asyncio.FIRST_COMPLETED)`.

```
Task was destroyed but it is pending!
task: <Task pending name='Task-891205' coro=<Queue.get() running at /usr/local/lib/python3.9/asyncio/queues.py:166> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f3a938c2310>()]>>
```

### DIFF
--- a/aiopg/utils.py
+++ b/aiopg/utils.py
@@ -137,27 +137,41 @@ class ClosableQueue:
     are consumed.
     """
 
-    def __init__(self, queue: asyncio.Queue):
+    __slots__ = ("_loop", "_queue", "_close_event")
+
+    def __init__(self, queue: asyncio.Queue, loop: asyncio.AbstractEventLoop):
+        self._loop = loop
         self._queue = queue
-        self._close_exception = asyncio.Future()  # type: asyncio.Future[None]
+        self._close_event = loop.create_future()
 
     def close(self, exception: Exception) -> None:
-        if not self._close_exception.done():
-            self._close_exception.set_exception(exception)
+        if self._close_event.done():
+            return
+        self._close_event.set_exception(exception)
 
     async def get(self) -> Any:
-        loop = get_running_loop()
-        get = loop.create_task(self._queue.get())
+        if self._close_event.done():
+            try:
+                return self._queue.get_nowait()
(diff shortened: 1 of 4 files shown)


tests/test_connection.py
```diff
@@ -595,37 +596,38 @@ async def test_connection_on_server_restart(connect, pg_server, docker):
...
-    docker.restart(container=pg_server["Id"])
 
-    try:
-        with pytest.raises(psycopg2.OperationalError):
-            await asyncio.wait_for(reader, 10)
-    finally:
-        conn.close()
-        reader.cancel()
-
-        # Wait for postgres to be up and running again before moving on
```
Pliner, line 619: I changed a restart of a container to network disconnection, just to simplify the test.

aiopg/utils.py
```diff
@@ -137,27 +137,41 @@ class ClosableQueue:
...
+        if self._close_event.done():
+            return
+        self._close_event.set_exception(exception)
 
     async def get(self) -> Any:
-        loop = get_running_loop()
-        get = loop.create_task(self._queue.get())
+        if self._close_event.done():
+            try:
+                return self._queue.get_nowait()
```
spaceorc, line 155: So, we extracting only one event. And what if there are many? Maybe skip this totally?

spaceorc (approved):

PR #865, opened 2021-07-08
Update PostgreSQL versions used in tests

and-semakin:
## What do these changes do?

Tests were using outdated/deprecated PostgreSQL versions. For better results tests should run over an actual PostgreSQL version.

I also had to remove some assertions in tests since they decided to phase out `oids` feature since PostgreSQL 12. See: https://stackoverflow.com/a/57022291/10650942

No. Only in tests.

## Checklist

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,9 @@
+1.3.2 (2021-07-XX)
+^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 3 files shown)


and-semakin:
The PR template is a little bit outdated. It mentions some `CHANGES` folder which I don't see anymore.

Pliner:
Thanks for the contribution.

PR #868, opened 2021-07-09
Run linters in CI

and-semakin:
## What do these changes do?

The project seems to use `flake8`, `mypy`, `black` & `isort` but doesn't enforce it in the CI. This PR makes it necessary for these tools to pass successfully.

I had to `# type: ignore` one type errors. At the moment I don't have a clear understanding of how to fix this issue better. The PR still adds some value because all other parts of the project will be type-checked, and no new type violations will be allowed.

No. This PR affects only maintainers and future contributors.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,19 +42,39 @@ jobs:
         python setup.py check -rms
         codecov
 
+  lint:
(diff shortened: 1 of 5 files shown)


Pliner:
Thanks!

PR #878, opened 2021-08-06
add py.typed marker, to indicate this package provides type hints

gjcarneiro:
## What do these changes do?

This package seems to to have type hints, but it is not signalling that it provides them.  This adds the signal.

See PEP 561.

When doing type checking, aiopg is now consulted for typing information.

## Checklist


asvetlov (approved):

aklim007:
this library does not provide type hints completely. and now it broke Mypy

` error: Call to untyped function "begin" in typed context  [no-untyped-call]`
` error: Call to untyped function "execute" in typed context  [no-untyped-call]`

PR #881, opened 2021-09-29
repeir error NotImplemented for Windows OS

eshikvtumane:
## What do these changes do?

Repair error NotImplemented for Windows OS and Python 3.8

No

## Checklist

- [ ] Unit tests for the changes exist (I don't know how check Windows OS in CI)
- [ ] Documentation reflects the changes (don't need)

### DIFF
--- a/aiopg/utils.py
+++ b/aiopg/utils.py
@@ -14,7 +14,10 @@ from typing import (
     Union,
 )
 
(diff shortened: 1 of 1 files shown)


Bonifatius94:
I have the same issues with Python 3.11. I think it's a more general issue than just Python 3.8
But interesting that it's not merged after 3 years.

PR #914, opened 2025-03-04
Don't depend on async-timeout for Python 3.11+

carlsmedstad:
This library has effectively been upstreamed into Python 3.11+ and is no longer actively supported.

See: https://archlinux.org/todo/deprecate-python-async-timeout/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ from pathlib import Path
 
 from setuptools import setup, find_packages
 
-install_requires = ["psycopg2-binary>=2.9.5", "async_timeout>=3.0,<5.0"]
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -3,7 +3,7 @@
 
 from setuptools import setup, find_packages
 
-install_requires = ["psycopg2-binary>=2.9.5", "async_timeout>=3.0,<5.0"]
+install_requires = ["psycopg2-binary>=2.9.5", "async_timeout>=3.0,<5.0; python_version < \"3.11\""]
```
webknjaz:
```suggestion
install_requires = ["psycopg2-binary>=2.9.5", "async_timeout>=3.0,<5.0; python_version < '3.11'"]
```

